PR #1, opened 2016-04-05
Fail loudly when 'text' key is missing.

benjaoming:
No one especially me would like to know how much time this took to figure out :)


benjaoming:
Sorry, this PR was meant as a kind of sketch of the intention. But I take that there are other uses of message dicts that do not need the 'text' key. My understanding of ASGI is far too little for me to fix this problem, but I guess if there's a way to make it fail loudly when supposedly mandatory elements of the message dict are missing, that'd be good.

It could also be that different message type object classes and constructors would be the way forward?

benjaoming:
This belongs in Daphne and has been fixed by @andrewgodwin -- thanks!

PR #5, opened 2016-09-05
Channel and group name error enchancement

jtiki:
Fixes issue stated in [Django Channels ticket 336](https://github.com/andrewgodwin/channels/issues/336)


andrewgodwin:
Looks like this is failing the conformance test suite - can you figure out why? I can't merge it if it breaks all the tests :)

jtiki:
I would but I keep getting "The repository at andrewgodwin/asgiref was not found."

charettes:
@jtiki the repository was moved under the Django project umbrella (`andrewgodwin/asgiref` -> `django/asgiref`). You'll have to reconfigure your remotes to point to the new origins.

jtiki:
@charettes I should have known that, I read the email this morning.

jtiki:
@andrewgodwin I just ran the tests on my machine but none of them failed:
screen shot 2016-09-10 at 7 40 01 pm

And just to make sure I wasn't doing the tests wrong, I ran all the tests manually:
screen shot 2016-09-10 at 7 44 22 pm

Everything checks out correctly in Py3.5. So I am not entirely sure what is causing this to fail on TravisCI.

andrewgodwin:
I just checked out this PR and ran them on my machine under Python 3.5, and they do all fail, so there's something weird going on here. What OS are you running them under? OSX? (I'm running Windows Subsystem for Linux here, so basically Ubuntu)

jtiki:
I'm running OSX 10.11.3  (El Capitan) with Python 3.5.0 installed via homebrew.

andrewgodwin:
It might be an OS difference then. I can't really do much about it at the moment, I might have a chance to try and debug next week.

jtiki:
Everything is good now, sorry about all the trouble getting this together for such a small change. I had to change some of the test files to look for an exception instead of the function returning False.

andrewgodwin:
Ahh, it was the pyc files being checked in, I should have caught that. Thanks!

PR #6, opened 2016-11-01
Altert Channel.send on the inmemory layer to send a copy of the message.

ostcar:
The following code acts different with the inmemory layer and the redis layer

```python
message = {'foo': [1,2,3]}
Channel('some_channel').send(message)
message['foo'][0] = 'new value'
```
The inmemory layer sends the object ```message ``` to the receiver and therefore the inner list as reverence. The redis layer creates a string from the message at the moment ```Channel(...).send()``` is called and therefore a copy of the message.

This pull request fixes this and both layers should work the same.

### DIFF
--- a/asgiref/conformance.py
+++ b/asgiref/conformance.py
@@ -266,3 +266,15 @@ class ConformanceTestCase(unittest.TestCase):
         """
         self.assertTrue(hasattr(self.channel_layer, "MessageTooLarge"))
         self.assertTrue(hasattr(self.channel_layer, "ChannelFull"))
+
+    def test_message_alteration_after_send(self):
+        """
+        Tests that a message can be altert after it was send through a channel
+        without affecting the object inside the queue.
+        """
+        message = {'value': [1, 2, 3]}
+        self.channel_layer.send('channel', message)
+        message['value'][0] = 'new value'
+
+        _, message = self.channel_layer.receive_many(['channel'])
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Couple of questions:

* Do you know what sort of performance difference this entails? Copy can be reasonably expensive, and it would be nice to know if we should offer shallow/no copy as an option.
* Could you move the no-modification test to the conformance test suite instead? I think all backends should pass this test.

ostcar:
I have no knowledge about the speed of deepcopy. It sounds slower then json: http://stackoverflow.com/questions/24756712/deepcopy-is-extremely-slow/29385667#29385667

In the redis backend you use msgpack to serialize the data. What do you think about 
```python
message = msgpack.unpackb(msgpack.packb(message, use_bin_type=True), encoding="utf8")
```
So the inmemory backend has the same speed the redis backend and it also makes sure, then the same data structures can be send to either of the backends.

Of cause I can move the test, I will do it after you make a decision with the deepcopy problem.

andrewgodwin:
An actual speed comparison would be what I would do, but I also don't want to pull in msgpack as a dependency just to try and solve this. If you can move the test into the conformance suite I'll pull it in with the deepcopy - I'd rather it be right than slightly faster.

ostcar:
Done

andrewgodwin:
Thanks!

PR #7, opened 2016-12-02
Update InMemory ChannelLayer group_send to call group_channels

linuxlewis:
group_channels is the defined method to use in the asgi spec


andrewgodwin:
I'm going to reject this as the code that's currently there is faster; there's no need to call the defined interface function as it's internal implementation details of itself.

PR #8, opened 2016-12-13
validation method for multiple  channel names

navyad:
added validation method that accepts list of channel names.

### DIFF
--- a/asgiref/base_layer.py
+++ b/asgiref/base_layer.py
@@ -112,3 +112,11 @@ class BaseChannelLayer(object):
             if bool(self.group_name_regex.match(name)):
                 return True
         raise TypeError("Group name must be a valid unicode string containing only alphanumerics, hyphens, or periods.")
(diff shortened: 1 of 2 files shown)


asgiref/base_layer.py
```diff
@@ -112,3 +112,11 @@ def valid_group_name(self, name):
             if bool(self.group_name_regex.match(name)):
                 return True
         raise TypeError("Group name must be a valid unicode string containing only alphanumerics, hyphens, or periods.")
+
+    def valid_channel_names(self, names):
+        _non_empty_list = True if names else False
+        _names_type = isinstance(names, list)
```
andrewgodwin: Yes, this test is quite strict, but the spec does say it should be a list when passed in; I'd be concerned about re-use of a generator more than once in the body. The alternative would be `not a string type and iterable` and then consume it into a list immediately.

PR #9, opened 2016-12-16
Add missed abstract method.

proofit404:
This method used in all already implemented backends. Also it's required in the conformance test. I think it worth to add it in the base class.


andrewgodwin:
Agreed!

PR #13, opened 2017-01-24
tox, and all the Pythons 

maikhoepfel:
This pull requests introduced tox to be able to test against multiple Python versions locally, and configures tox and Travis so that they test against all Python versions supported by Channels. I also added a bit of documentation and cleanup.

I intend to make similar changes to the other Channels projects, so this pull request is a good occasion to discuss whether the same changes can and should be made to the other projects.

### DIFF
--- a/asgiref/conformance.py
+++ b/asgiref/conformance.py
@@ -29,22 +29,10 @@ class ConformanceTestCase(unittest.TestCase):
     capacity_limit = None
 
     @classmethod
-    def raiseSkip(cls, message):
-        """
-        Picks the right skip class based on what version of unittest we're using.
-        """
-        try:
-            import unittest2
-            SkipTest = unittest2.SkipTest
-        except ImportError:
-            SkipTest = unittest.SkipTest
-        raise SkipTest(message)
-
-    @classmethod
     def setUpClass(cls):
         # Don't let this actual class run, it's abstract
         if cls is ConformanceTestCase:
-            cls.raiseSkip("Skipping base class tests")
+            raise unittest.SkipTest("Skipping base class tests")
 
     def setUp(self):
         if self.channel_layer is None:
@@ -60,7 +48,7 @@ class ConformanceTestCase(unittest.TestCase):
         We can't use the decorators, as we need access to self.
         """
         if extension not in self.channel_layer.extensions:
-            self.raiseSkip("No %s extension" % extension)
+            raise unittest.SkipTest("No %s extension" % extension)
 
     def test_send_recv(self):
         """
@@ -266,7 +254,7 @@ class ConformanceTestCase(unittest.TestCase):
         after the right number of messages. Only runs if capacity_limit is set.
         """
         if self.capacity_limit is None:
-            self.raiseSkip("No test capacity specified")
+            raise unittest.SkipTest("No test capacity specified")
         for _ in range(self.capacity_limit):
             self.channel_layer.send("cap_test", {"hey": "there"})
         with self.assertRaises(self.channel_layer.ChannelFull):
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ dist/
 build/
 __pycache__/
 __pycache__/*
+/.tox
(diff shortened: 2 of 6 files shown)


proofit404:
Conformanse test suite should not be moved to the tests package. It used in the packages like asgi_redis and asgi_rabbitmq as regular module. Also I think tests package should be excluded from PyPI package.

maikhoepfel:
@proofit404 Thank you for the quick feedback. I see your point about `conformance` being used as a base for tests elsewhere; I wasn't aware of that, sorry. I suppose then technically we need a split between "testing support" and "tests", with conformance belonging in the first category. Similar to how `django` makes a distinction between `/tests` and `django.test`. But it might be a bit early for that, so I'm happy to just leave it where it is, accepting that then testing related code is still in the same module as the production code.

What do you think of the other changes? 

> Also I think tests package should be excluded from PyPI package.
I'd rather not address packaging related issues as part of this pull request.

proofit404:
We have some package related discussion in the https://github.com/django/channels/pull/497 Feel free to join.

In the `requirements_dev.txt` you install tox and virtualenv. But this dependencies not used on Travis CI since you run unittest directly. Then you install dev requirements **inside** tox virtualenv where they are unnecessary as well.

`unittest2` PyPI package is a backport of python2.7 unittest for earlier python versions. Since channels support python2.7 only I agree with this change.

maikhoepfel:
I agree with your suggestion in https://github.com/django/channels/pull/497. I have reverted my move of the conformance tests; I think it makes sense to address it as part of a wider "project layout" round of PRs.

You're right that `requirements_dev.txt` isn't necessary inside tox nor Travis - good catch, thanks. I tend to use pytest (and did originally for this PR), where it would be necessary, and hence forgot to remove it here.

asgiref/conformance.py
```diff
@@ -29,22 +29,10 @@ class ConformanceTestCase(unittest.TestCase):
...
-        except ImportError:
-            SkipTest = unittest.SkipTest
-        raise SkipTest(message)
-
-    @classmethod
     def setUpClass(cls):
         # Don't let this actual class run, it's abstract
         if cls is ConformanceTestCase:
-            cls.raiseSkip("Skipping base class tests")
+            raise unittest.SkipTest("Skipping base class tests")
```
andrewgodwin: Weird, not doing this made the conformance tests not skip correctly on Python 2.

andrewgodwin:
I think we should drop Python 3.2 and 3.3 support (Django only supports 3.5 as of 2.0 after all, but I'm OK with 3.4 too). This means that we can also drop the requirements_dev stuff.

asgiref/conformance.py
maikhoepfel: I'm afraid I don't understand; could you please clarify? What do you mean by "not doing this"? Are the tests skipped correctly with my changes? As far as I can tell, with and without my changes, 20 tests are run on both Python 2 and 3.

maikhoepfel:
@andrewgodwin I'm happy to drop Python 3.2 and 3.3 support - it is a pragmatic choice. It would be nice to say we support everything that Django supports, but actual usage is low. I included 3.2 and 3.3 because I thought you had agreed to that when discussing the plans for the first milestone. I have removed it now, and will aim for 2.7, 3.4 and 3.5 support for the other Channels projects as well. 

I realize requirements_dev.txt isn't a pretty solution, but when working on django-oscar I did find it's important for newcomers to be able to set up a development/test environment for the project easily - and it's the best pattern that I've come across to achieve that. I agree it's not strictly necessary when it's just tox that is needed, but I presumed the environment might be more complex for the other Channels projects. Let's revisit it if the need arises; I've removed it for now. 

That only leaves the test skipping issue above. I've rebased my pull requests with the changes.

andrewgodwin:
Well the requirements_dev thing was mostly that Python 3.2 support was the only thing needing a requirements_dev, so by dropping that the dev requirements are satisfied by `pip install -e .`.

Looks good, I'll merge it in.

PR #14, opened 2017-02-07
Update base_layer.py

Nablyudatel:
Now only the English symbols are resolved. It is a problem. Unicode allows more.


andrewgodwin:
This is deliberate, channel names are limited to only valid ASCII character names for backend compatability reasons. Sorry.

Nablyudatel:
Its worked.

PR #19, opened 2017-05-16
Optionally retry on receive in ConformanceTest

bgreenberg-eb:
Added an option to `ConformanceTestCase` that allows the test case to retry receiving. This is necessary for testing `ChannelLayer` implementations that do not guarantee a response will be returned on every `receive()` call, even when there are messages on the channel. This would be the case, for example, for a channel layer designed for a multi-worker environment with multiple backing hosts, that checks a different host on each call.

Also fixed `test_send_recv` so that it does not depend on the order in which messages are returned, since order is not guaranteed in the spec.

### DIFF
--- a/asgiref/conformance.py
+++ b/asgiref/conformance.py
@@ -74,14 +74,16 @@ class ConformanceTestCase(unittest.TestCase):
         self.channel_layer.send("sr_test", {"value": "green"})
         self.channel_layer.send("sr_test", {"value": "yellow"})
         self.channel_layer.send("sr_test2", {"value": "red"})
-        # Receive from the first channel twice; order is not guaranteed
+        # Receive from the first channel twice
         response_messages = []
         for i in range(3):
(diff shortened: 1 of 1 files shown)


asgiref/conformance.py
```diff
@@ -27,6 +27,21 @@ class ConformanceTestCase(unittest.TestCase):
...
+        """
+        Allows tests to automatically call channel_layer.receive() more than once.
+        This is necessary for testing ChannelLayer implementations that do not guarantee a response will
+        be returned on every receive() call, even when there are messages on the channel. This would be
+        the case, for example, for a channel layer designed for a multi-worker environment with multiple
+        backing hosts, that checks a different host on each call.
+        """
+        for _ in range(self.receive_tries):
+            channel, message = self.channel_layer.receive(channels)
+            if channel:
```
andrewgodwin: Can we tighten this up to `if channel is not None`?

PR #20, opened 2017-06-03
Verify async extension against its methods.

proofit404:
Just to be more precise.

Should we do the same for `twisted` extension?


andrewgodwin:
Hmm, I didn't do this as the `receive_async` test implicitly checks that the method exists. Did you run into a case where that didn't work?

proofit404:
On Python 2 we don't have `receive_async` method, but extensions list contains `async` extension anyway.

andrewgodwin:
Ah, right, and the test turns itself off on Python 2! I'll merge it in, thanks.

PR #21, opened 2017-07-17
Add classifiers to setup.py

dsanders11:
Without language classifiers in `setup.py` this project gets marked as not supporting Python 3 when using a tool like [caniusepython3.com](https://caniusepython3.com) to check Python 3 compatibility. I copied the classifiers from `channels` as they seemed applicable.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,4 +25,19 @@ setup(
     entry_points={'console_scripts': [
         'asgiref_benchmark = asgiref.benchmark:Benchmarker.cli',
     ]},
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Right! I always forget these.

dsanders11:
@andrewgodwin, this appears to have regressed, they are no longer in `setup.py`.

andrewgodwin:
Weird... must have been a bad merge. Added them back.

PR #23, opened 2017-08-21
Improve channel and group name validation

jaywink:
Got hit by a situation where my group name was over 100 characters but the system was saying it contained illegal characters, which was a bit confusing. Trying to improve the messages coming out of the validation with these commits.

* Functions valid_channel_name and valid_group_name were using almost the same error message. Unified both to use the class 'invalid_name_error' attribute. This also added to the error message the underscore as a valid character. It was a valid character all the time but was missing from the raised error.

* Refactor BaseChannelLayer.match_type_and_length. Channel and group names are also restricted in length and type, but if those fail to validate we got a validation message related to characters in the name. Raise a separate TypeError to indicate clearly whether it was the length or the type that failed validation.

Thanks for making Channels!

### DIFF
--- a/asgiref/base_layer.py
+++ b/asgiref/base_layer.py
@@ -93,10 +93,12 @@ class BaseChannelLayer(object):
                 return capacity
         return self.capacity
 
-    def match_type_and_length(self, name):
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Thanks!

PR #34, opened 2018-02-25
Fixed #33: Clarify that the `websocket` protocol does not expose frag…

### LINKED ISSUE #33: WebSocket spec speaks of "data frame" but would be better to use "message"
In `www.rst`, WebSocket Send and Receive sections speak of "data frames": https://github.com/django/asgiref/blob/dfb76d4e23d0851021ee44529e90bac2453b50e4/specs/www.rst#receive

In WebSockets (as you probably know) there is a distinction between frames and messages: a message consists of one or more frames. Or more precisely it can be said that a complete data frame (= message) can be fragmented to multiple incomplete data frames.

The Daphne server is using messages, and does not expose details of fragmentation. So I assume the spec want to use messages too.

---

Some discussion on whether the spec should actually expose fragmentation.

Exposing frames to the application can be useful for implementing streaming messages. For example, I might want to send some huge file over a binary WebSocket message; the sending side breaks the message up to 64kb frames, and the received side appends them to the disk, thus there is no need to buffer the entire file in memory.

The obvious way to expose fragmentation in ASGI is to use a similar convention to the http `more_body` field - `more_data: True` if there are more frames to come, `more_data: False` for the last frame. However there is a little technicality which makes this difficult. From [RFC 6555](https://tools.ietf.org/html/rfc6455#section-5.6):

> Note that a particular text frame might include a partial UTF-8 sequence; however, the whole message MUST contain valid UTF-8.

i.e. it is not possible to deliver text frames as `str` (as opposed to `bytes`), because it might not decode at the boundary.

Another way too enable streaming is to keep `more_data` but do it on arbitrarily-sized chunks and disregard frames.

But overall, it seems pretty clear to me that the 99% use case is to deliver full buffered messages, and that the spec should replace frames -> messages.

Note:

- Many servers buffer frames *anyway*.
- The original JavaScript `WebSocket` API doesn't expose frames.

bluetech:
…mentation

It is up to the ASGI server to assemble the fragments and deliver only
complete messages to the application. To make it clear, replace mentions
of "frames" with "messages", according to the terminology of the RFC:

    After a successful handshake, clients and servers transfer data back
    and forth in conceptual units referred to in this specification as
    "messages".  On the wire, a message is composed of one or more
    frames. [...]

    A frame has an associated type.  Each frame belonging to the same
    message contains the same type of data.

https://tools.ietf.org/html/rfc6455#section-1.2

### DIFF
--- a/specs/www.rst
+++ b/specs/www.rst
@@ -166,8 +166,11 @@ WebSockets share some HTTP details - they have a path and headers - but also
 have more state. Again, most of that state is in the scope, which will live
 as long as the socket does.
 
-WebSocket protocol servers should handle PING/PONG requests themselves, and
-send PING frames as necessary to ensure the connection is alive.
+WebSocket protocol servers should handle PING/PONG messages themselves, and
+send PING messages as necessary to ensure the connection is alive.
+
+WebSocket protocol servers should handle message fragmentation themselves,
+and deliver complete messages to the application.
 
 The WebSocket protocol should be signified to ASGI applications with
 a ``type`` value of ``websocket``.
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Great, thanks!

Issue #32, opened 2018-02-25
Assert that there is no message to receive using ApplicationCommunicator


dgilge:
What do you think about having a method to assert that there is no message to receive using the `ApplicationCommunicator`?

Right now I'm doing it like this (using Channels' `WebsocketCommunicator`): 

```python
    with pytest.raises(futures.TimeoutError):
        await communicator.receive_from()
```

I'd like to have something like

```python
    assert communicator.new_messages(timeout=1) == 0
    # or maybe
    assert communicator.received_nothing(timeout=1)
```

Actually the `timeout` should be 0 by default. Otherwise the tests might take quite long if you have a lot of these checks.

andrewgodwin:
Yes, we should probably have this - I won't be able to add it myself for a while though as I've got bugs to fix elsewhere. Feel free to submit a PR if you want.

dgilge:
Ok, fine. What about the following? (I'm not sure about the sleep times yet. And I'd actually prefer to implement `queue_empty` simpler...)

```python
    async def queue_empty(self, timeout=0.1):
        if not self.output_queue.empty():
            return False
        if timeout <= 0.01:
            await asyncio.sleep(timeout)
            return self.output_queue.empty()
        await asyncio.sleep(0.01)
        if not self.output_queue.empty():
            return False
        await asyncio.sleep(timeout - 0.01)
        return self.output_queue.empty()

    async def queue_count(self, wait=0.1):
        await asyncio.sleep(wait)
        return self.output_queue.qsize()
```

andrewgodwin:
I'd just have one top-level function called `communicator.receive_nothing` that raises an (assertion?) error if it _does_ receive something, that just wraps receive_from and catches the `TimeoutError`.

dgilge:
I also thought about doing it that way. But as far as I understand the communicator is closed after this exception which is not what I want in some use cases...

andrewgodwin:
Ah yes, the timeout will cancel it. I still thing calling it `receive_nothing` makes sense though, and implement inside that as you see fit (using queue empty detection is fine)

PR #37, opened 2018-03-12
ApplicationCommunicator's receive_nothing

dgilge:
Addresses #32.

On my computer it took < 0.001 s to send a message via (Channels') `group_send` using the `InMemoryChannelLayer`. Therefore an attempt after 0.01 s should actually be enough. Please let me know if you prefered a simpler implementation.

**ToDo**

- [x] tests (or not necessary? – There are no tests for testing.)
- [x] docs (in Channels repo)

### DIFF
--- a/tests/test_testing.py
+++ b/tests/test_testing.py
@@ -1,0 +1,67 @@
+import math
+import time
+
+import pytest
+
+from asgiref.testing import ApplicationCommunicator
+from asgiref.wsgi import WsgiToAsgi
+
+
+@pytest.mark.parametrize(
+    "event, timeout, attempt, duration, tolerance, expected", [
+        # No event
+        # Standard values
+        (False, 0.1, 0.01, 0.1, 0.03, True),
+        # attempt > timeout
+        (False, 0.1, 1, 0.1, 0.03, True),
+        # timeout % attempt > 0
+        (False, 1, 0.6, 1, 0.03, True),
+
+        # Events
+        # Standard values
+        (True, 0.1, 0.01, 0.01, 0.003, False),
+        # No delay
+        (True, 0, 0.01, 0, 0.003, True),
+        # No attempt
+        (True, 0.1, 0, 0.1, 0.03, False),
+        # attempt > timeout
+        (True, 0.1, 1, 0.1, 0.03, False),
+    ]
+)
+@pytest.mark.asyncio
+async def test_receive_nothing(event, timeout, attempt, duration, tolerance, expected):
+    """
+    Tests ApplicationCommunicator.receive_nothing with different parameters.
+    """
+    # Get an ApplicationCommunicator instance
+    def wsgi_application(environ, start_response):
+        start_response("200 OK", [])
+        yield b"content"
+    application = WsgiToAsgi(wsgi_application)
+    instance = ApplicationCommunicator(application, {
+        "type": "http",
+        "http_version": "1.0",
+        "method": "GET",
+        "path": "/foo/",
+        "query_string": b"bar=baz",
+        "headers": [],
+    })
+
+    if event:
+        await instance.send_input({
+            "type": "http.request",
+        })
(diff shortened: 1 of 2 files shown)


andrewgodwin:
I would expect `attempt` to be the time delay inside a `while` loop capped by `timeout`, to be honest, not to just be a two-delay function. I think rewriting as a loop will make the code a lot simpler (no need for the special first case, if you check before the delay).

I'd also like to see this added to the documentation, and a test in the `test_testing` file.

dgilge:
Thanks for your suggestions. I like the idea with the loop. Unfortunately, I didn't manage to make it a lot simpler. What I could do is `assert 0 < attempt < timeout` inside `receive_nothing` but I don't like this limitation. Right now there will just be no attempt until the timeout if this assertion isn't true.

Here's another solution with a `while` loop but I don't like it so much. It seems to be slower and as a result one test failed (the one with no delay). (I didn't test it in every detail.)

```python
    async def receive_nothing(self, timeout=0.1, attempt=0.01):
        """
        Checks that there is no message to receive in the given time.
        """
        if timeout > attempt:
            duration = 0
            while duration < timeout:
                if not self.output_queue.empty():
                    return False
                await asyncio.sleep(attempt)
                duration += attempt
                if attempt <= 0:
                    break
            await asyncio.sleep(timeout - (duration - attempt if attempt > 0 else 0))
        else:
            await asyncio.sleep(timeout)
        return self.output_queue.empty()
```

Sorry when I'm not so sure how to implement it nice and when I have many questions. I'm kind of a beginner. ;)

I'll add it to the Channels docs when we are more sure about the implementation.

andrewgodwin:
I'd recommend code more like this (I find it best not to rely on sleeps being exactly the right length of time in asyncio, and I have changed `attempt` to be called `interval`):

```
start = time.monotonic()
while time.monotonic() - start < timeout:
    if queue is not empty:
        return False
    await sleep(interval)
return (queue is empty)
```

dgilge:
I've updated the PR according to your recommendation.

tests/test_testing.py
```diff
@@ -0,0 +1,67 @@
...
+        # interval > timeout (interval has precedence over timeout)
+        (False, 0.1, 1, 1, 0.03, True),
+        # timeout % interval > 0 (interval cycle completes)
+        (False, 1, 0.6, 1.2, 0.03, True),
+
+        # Events
+        # Standard values
+        (True, 0.1, 0.01, 0.01, 0.003, False),
+        # No delay
+        (True, 0, 0.01, 0, 0.003, True),
```
dgilge: I'm not sure whether to keep this test or not because it might fail in some environments (the message might be in the queue already).

andrewgodwin: Don't keep it like this - there is no need to be this precise. All you need to do is test that it returns `True` or `False` correctly and within a simple overall timeout.

dgilge: I thought so. However, I wanted to make sure that the `reveice_nothing` behaves as I supposed in the first place. - I've updated the test.

tests/test_testing.py
```diff
@@ -0,0 +1,39 @@
...
+    })
+
+    # No event
+    assert await instance.receive_nothing() is True
+
+    # Events
+    await instance.send_input({
+        "type": "http.request",
+    })
+    assert await instance.receive_nothing() is False
```
andrewgodwin: Can you put a comment above each `receive_nothing`/`receive_output` pair to say what part of the response they're receiving?

dgilge:
@andrewgodwin 
I added the comments to the test and opened a PR for the docs on Channels.

Thank you for your guidance! Sometimes I thought the code somehow should look nicer and your comments gave me the hints I needed.

andrewgodwin:
Thanks for the PR! And I'm always happy to help make code look nicer :)

PR #40, opened 2018-05-12
Bump async-timeout version to 3.0

ttill:
Resolves dependency conflict with aiohttp.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ setup(
         ],
     },
     install_requires=[
-        'async-timeout~=2.0',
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Thanks!

adamhooper:
Does this warrant a fresh release? Until release, `aiohttp` and `channels` conflict.

andrewgodwin:
I just pushed a `2.3.1` release into our pipeline with this change.

adamhooper:
@andrewgodwin Thank you! I know this wasn't as simple as it ought to have been; but in the end it totally works :).

PR #42, opened 2018-05-15
WWW Spec changes

pgjones:
## WWW Spec: Allow for a HTTP response before Websocket handshake acceptance

This allows HTTP responses such as 401, authentication required to be
sent to the client by a framework that requires authenticated
websocket connections. It would also allow for redirect responses or
other more exotic framework responses.

## WWW Spec: Add a server push message

This specifies how a client should indicate that a HTTP/2 server push
should be initiated.

### DIFF
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -1,0 +1,41 @@
+Extensions
+==========
+
+The ASGI specification provides for server-specific extensions to be
+used outside of the core ASGI specification. This document specifies
+some common extensions.
+
+
+Websocket Denial Response
+-------------------------
+
+Websocket connections start with the user agent sending a HTTP request
+containing the appropriate upgrade headers. On receipt of this request
+a server can choose to either upgrade the connection or respond with a
+HTTP response (denying the upgrade). The core ASGI specification does
+not allow for any control over the denial-response, instead specifying
+that the HTTP status code ``403`` should be returned, whereas this
+extension allows an ASGI framework to control the
+denial-response. This is an extension, rather than a core part of
+ASGI, as most user agents do not utilise the denial response and hence
+this is a niche feature.
+
+ASGI Servers that implement this extension will provide
+``websocket.http.response`` in the extensions part of the scope::
+
+    "scope": {
+        ...
+        "extensions": {
+            "websocket.http.response": {},
+        },
+    }
+
+This will allow the ASGI Framework to send HTTP response messages
+after the ``websocket.connect`` message. These messages cannot be
+followed by any other websocket messages as the server should send a
+HTTP response and then close the connection.
+
+The messages themselves should be ``websocket.http.response.start``
+and ``websocket.http.response.body`` with a structure that matches the
+``http.response.start`` and ``http.response.body`` messages defined in
+the HTTP part of the core ASGI specification.
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -39,3 +39,37 @@ The messages themselves should be ``websocket.http.response.start``
 and ``websocket.http.response.body`` with a structure that matches the
 ``http.response.start`` and ``http.response.body`` messages defined in
 the HTTP part of the core ASGI specification.
+
+HTTP/2 Server Push
+------------------
+
+HTTP/2 allows for a server to push a resource to a client by sending a
+push promise. ASGI servers that implement this extension will provide
+``http.response.push`` in the extensions part of the scope::
+
+    "scope": {
+        ...
+        "extensions": {
+            "http.response.push": {},
+        },
+    }
+
+An ASGI framework can initiate a server push by sending a message with
+the following keys, This message can be sent at any time after the
+*Response Start* message but before the final *Response Body* message.
+
+Keys:
+
+* ``type``: ``http.response.push``
+
+* ``path``: Unicode string HTTP path from URL, with percent escapes
+  decoded and UTF-8 byte sequences decoded into characters.
+
+* ``headers``: A list of ``[name, value]`` lists, where ``name`` is the
+  byte string header name, and ``value`` is the byte string
+  header value. Header names must be lowercased.
+
+The ASGI server should then attempt to send a server push (or push
+promise) to the client. If the client accepts the server should create
+a new connection to a new instance of the application and treat it as
+if the client had made a request.
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -14,4 +14,4 @@ ASGI standard, as well as other examples and tips on implementation.
 
    introduction
    specs/index
-
+   extensions


specs/www.rst
```diff
@@ -149,6 +149,29 @@ Keys:
   will be ignored. Optional, defaults to ``False``.
 
 
+Server Push
+'''''''''''
+
+Indicates that the server should send a server push (or push promise)
+to the client. This message can be sent at any time after the
+*Response Start* message but before the final *Response Body* message.
```
andrewgodwin: So, Server Push responses need a body if I recall correctly; are you proposing this to be the type of push where the application sends back what it wants pushed and then the server calls itself with the given path and headers to get the content?

specs/www.rst
```diff
@@ -149,6 +149,29 @@ Keys:
...
+* ``path``: Unicode string HTTP path from URL, with percent escapes
+  decoded and UTF-8 byte sequences decoded into characters.
+
+* ``headers``: A list of ``[name, value]`` lists, where ``name`` is the
+  byte string header name, and ``value`` is the byte string
+  header value. Header names must be lowercased.
+
+Note:
+
+If the ``http_version`` is not 2 this message should be ignored.
```
andrewgodwin: Should probably just start saying these are hard errors?

specs/www.rst
```diff
@@ -223,12 +246,16 @@ Connection
...
+This message must be responded to with either an *Accept* message, a
+*Response Start* message, or a *Close* message. If a Response Start
+message is sent it should be followed by *Response Body* messages as
+required by the HTTP spec. Websocket messages cannot be sent after a
+*Response Start* message. If an *Accept* message is sent the handshake
+is completed and the socket will pass ``websocket.receive``
+messages. The protocol server must send an accept message during the
+handshake phase of the WebSocket and not complete the handshake until
+it gets a reply, returning HTTP status code ``403`` if the connection
+is denied.
```
andrewgodwin: I guess it's fine to have the websocket and http events intermingle, they're in the same document for a reason. Could you specifically say "HTTP Response Start" etc. here to make that clear?

specs/www.rst
pgjones: Yep, I've updated to hopefully make this clear.

specs/www.rst
pgjones: I've done so.

andrewgodwin:
OK so I largely agree with this but I need a bit more convincing as I have learnt to give myself a reasonably high bar for PRs against specs, so I would like to debate this a little (please humour me!)

The reason I initially did not have this in there was because I didn't see a large need for responding to an incoming WebSocket connection with a detailed HTTP response; in particular, any WebSocket client is not really capable of showing any received HTML/response body so only the HTTP error code makes sense for an end-user to provide (and in many clients, especially browser ones, even this is hard to get at).

Additionally, the only use of headers that RFC 6455 even mentions in an HTTP error response is for a WebSocket version field, which the server would be handling, not an application, so there appears to be no need for headers in a handshake error response, either.

Given these, why allow a full HTTP response? Should we instead be allowing HTTP error codes in the `websocket.close` message you send to deny an incoming handshake?

andrewgodwin:
Hmm, right, that situation makes sense, though it's still a bit strange. The question I'm asking is if it's worth asking servers to implement a LOT more complexity (in Daphne's case, this is going to require a lot of new code as the HTTP and WebSocket paths are very separate) for a very edge use-case.

ASGI already deliberately does not cover some of the more obscure parts of the HTTP spec - like sending headers mid-stream between chunks - because it's not worth supporting a 1% feature to make code more complex. Do you think this is? Would having it not be there be a make-or-break for you to help adopt it? Once something is in a spec it is near-impossible to remove.

Also, as for close codes, I agree mixing the namespace would be confusing. Could add a separate `http_status` option to the close instead to make it cleaner.

pgjones:
I see, I do agree it is a quite niche use case, however I don't feel I can remove the feature from Quart. I think it could be specified as an optional extension, I'm thinking something like:

```
"scope": {
    ...
    "extensions": {
        "websocket.http.response.start": {},
        "websocket.http.response.body": {},
    },
}
```
where the structure of these messages match the http.response equivalents, and the type just has the `websocket.` prefix. In addition I think there should be a _known_ extensions file to specify these and others. Does this sound good?

andrewgodwin:
Ah yes, an extension is probably the right choice here, and it can then still be bought into the spec proper if it proves useful and gets implemented quite widely. Fragmentation is never great, but in this case it's a small enough use case I'd be OK.

I would call the extension `websocket.http_denial`, or similar, though.

As for where to put them - maybe put an extensions section in the `docs/` folder and have known extensions there? That way they will end up on http://asgi.readthedocs.io, which I intend to use as the place to point people to for the spec.

docs/extensions.rst
```diff
@@ -0,0 +1,43 @@
...
+-------------------------
+
+Websocket connections start with the user agent sending a HTTP request
+containing the appropriate upgrade headers. On receipt of this request
+a server can choose to either upgrade the connection or respond with a
+HTTP response (denying the upgrade). The core ASGI specification does
+not allow for any control over the denial-response, instead specifying
+that the HTTP status code ``403`` should be returned, whereas this
+extension allows an ASGI framework to control the
+denial-resposne. This is an extension, rather than a core part of
```
andrewgodwin: Response misspelt.

docs/extensions.rst
```diff
@@ -0,0 +1,43 @@
...
+ASGI, as most user agents do not utilise the denial response and hence
+this is a niche feature.
+
+ASGI Servers that implement this extension will provide
+``websocket.http.response.start`` and ``websocket.http.response.body``
```
andrewgodwin: Let's have one key, not two. What would it mean if it just provided `websocket.http.response.body`? It wouldn't make sense. Also not a fan of that namespacing, maybe just have `websocket.something`.

specs/www.rst
```diff
@@ -149,6 +149,27 @@ Keys:
...
 
 
+Server Push
+'''''''''''
+
+Indicates that the server should send a server push (or push promise)
+to the client. This message can be sent at any time after the
+*Response Start* message but before the final *Response Body*
+message. The server should create a new connection if the client
+accepts the push and treat it as if the client had made a request.
```
andrewgodwin: Can HTTP/2 clients reject pushes back to the server? I thought it was always-send style and the client drops it if it doesn't want to use it.

andrewgodwin:
Left a couple of comments on the extensions thing and one on server push!

specs/www.rst
pgjones: I think the client can send a settings frame indicating that it won't accept server push frames.

pgjones:
Thanks, I've updated.

specs/www.rst
pgjones: Agreed, and done.

andrewgodwin:
One last thing I just thought about - are we requiring all servers to support HTTP/2 Server Push? If so, this would mean incrementing the version number on the specification.

If not, what should the behaviour be? Advertising availability via an extension? Just ignoring server push messages if they're not supported?

pgjones:
In the current scheme I think the server is expected to raise an Exception if it gets a message it doesn't understand, so I think that is the desired behaviour for existing servers. I think then this should mean an increment in the version number. To help with this I think the scope should gain an `asgi_version` parameter (that by default is 2.0) to compliment the extension look before you leap behaviour. 

Shall I add a commit to add this to the scope and increment the version to 2.1?

pgjones:
Ok, I've changed both to be extensions and we can see how servers adopt. I'd personally make h2 a core part of ASGI, but I don't think the community is there yet.

I'm not sure how best to add the asgi part given a version number bump isn't required. I agree with the structure though.

andrewgodwin:
Finally merged! Thanks for your patience.

pgjones:
Great, thanks :)

PR #45, opened 2018-06-02
Add startup/cleanup ability, bump the version and add asgi versioning

pgjones:
This is an indication of what I'm using in [Quart](https://gitlab.com/pgjones/quart) and [Hypercon](https://gitlab.com/pgjones/hypercorn). It would be great to get naming/feature feedback.

### DIFF
--- a/specs/asgi.rst
+++ b/specs/asgi.rst
@@ -2,7 +2,7 @@
 ASGI (Asynchronous Server Gateway Interface) Specification
 ==========================================================
 
-**Version**: 2.0 (2017-11-28)
+**Version**: 2.1 (2018-06-03)
 
 Abstract
 ========
@@ -194,10 +194,14 @@ The application interface is specified as the more generic case of two callables
 to allow more flexibility for things like factory functions or type-based
 dispatchers.
 
-Both the ``scope`` and the format of the messages you send and receive are
-defined by one of the application protocols. ``scope`` must be a ``dict``.
-The key ``scope["type"]`` will always be present, and can be used to work
-out which protocol is incoming.
+Both the ``scope`` and the format of the messages you send and receive
+are defined by one of the application protocols. ``scope`` must be a
+``dict``.  The key ``scope["type"]`` will always be present, and can
+be used to work out which protocol is incoming. The key
+``scope["asgi"]`` will also be present as a dictionary containing a
+``scope["asgi"]["version"]`` key that corresponds to the ASGI version
+the server implements. As this was added in version 2.1, if missing
+the version should default to ``"2.0"``.
 
 The protocol-specific sub-specifications cover these scope
 and message formats. They are equivalent to the specification for keys in the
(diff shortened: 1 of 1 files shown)


specs/asgi.rst
```diff
@@ -179,14 +184,24 @@ implementations, as a class::
         async def __call__(self, receive, send):
             ...
 
+        async def startup(self):
```
andrewgodwin: You have written these as methods on `self`, but they would in fact need to be `classmethod`s, no? The application object isn't instantiated at server start.

specs/asgi.rst
```diff
@@ -179,14 +184,24 @@ implementations, as a class::
...
-The key ``scope["type"]`` will always be present, and can be used to work
-out which protocol is incoming.
+Both the ``scope`` and the format of the messages you send and receive
+are defined by one of the application protocols. ``scope`` must be a
+``dict``.  The key ``scope["type"]`` will always be present, and can
+be used to work out which protocol is incoming. The key
+``scope["asgi"]`` will also be present as a dictionary containing a
+``scope["asgi"]["version"]`` key that corresponds to the ASGI version
+the server implements. As this was added in version 2.1, if missing
+the version should default to ``"2.0"``.
```
andrewgodwin: Agree with this bit.

andrewgodwin:
If this is going to go in the spec it needs to be very clear about the cases where a server:

* Preforks (run once before the fork, or n times after?)
* Runs separate worker processes (run in the main process, or the workers?)

And also, should it always be run when the server exits with an error (if possible)?

specs/asgi.rst
pgjones: Yep, this is a silly mistake.

specs/asgi.rst
```diff
@@ -168,6 +168,11 @@ It must return another, awaitable callable::
 * ``send``, an awaitable callable taking a single event dict as a positional
   argument that will return once the send has been completed
 
+The application may have ``startup`` and ``cleanup`` attributes that
+are coroutine functions taking no arguments. If these attributes exist
+they will be awaited once per server process before serving a single
+request and after serving all requests respectively.
```
tomchristie: Is there any reason why startup and cleanup couldn't be handled as message types, rather than introducing extra API? (Eg. Use something like `events.startup`, `events.cleanup`, `events.tick`)

pgjones: I don't think they could be message types received via the receive function (as that would be after serving and the scope creation) but I think they could be different scope types. Using attributes was an existing suggestion for ASGI.

andrewgodwin: I feel a bit weird about doing it as scope types but maybe that would actually be a cleaner API, thinking about it. Would you launch one application instance for the entire server lifetime though, and send it one event at the start and one at the end?

tomchristie: That’s what I’d do, yup. No new API, just a new scope type.

andrewgodwin: OK, in that case I am +1 to that being the approach. @pgjones, what do you think?

pgjones:
Closed in favour of #50.

tomchristie:
Consideration: How this proposal would affect middleware.

PR #46, opened 2018-06-13
Update asgi.rst: websocket.send is missing

samzer:


andrewgodwin:
Thanks! That was an unfortunate repetition.

PR #49, opened 2018-06-24
Clarify backpressure handling on connection close

pgjones:
This makes it clear that the send awaitable callable should unblock
and act as a no-op after the connection has closed.


andrewgodwin:
Lovely, thanks.

PR #50, opened 2018-07-01
Add a Lifespan protocol, and add asgi versioning

pgjones:


pgjones:
@andrewgodwin @tomchristie as a protocol, rather than attribute, suggestion.

andrewgodwin:
Hmm does this belong in the main ASGI spec? I would think it's better off as its own protocol doc, like the `www.rst` one.

pgjones:
Not sure what you mean? It is in a `lifespan.rst` doc?

andrewgodwin:
Buh, tired eyes yesterday, sorry. Will review it properly now.

specs/asgi.rst
```diff
@@ -323,6 +329,7 @@ Version History
 ===============
 
 * 2.0 (2017-11-28): Initial non-channel-layer based ASGI spec
+* 2.1 (2018-07-01): Updated ASGI spec with a lifespan protocol.
```
andrewgodwin: I'm not sure that a new protocol needs a new core ASGI version, to be honest. I do think adding the version key is sensible, though, but maybe just at a 2.0 for now.

specs/lifespan.rst
```diff
@@ -0,0 +1,55 @@
+=================
+Lifespan Protocol
+=================
+
+**Version**: 2.1 (2018-07-01)
+
+The Lifespan ASGI sub-specification outlines how to communicate
+lifespan events such as startup and cleanup within ASGI. The lifespan
+being refered to is that of main event loop.
```
andrewgodwin: Can we get a clearer definition of what this means in a multi-process environment? E.g. a main coordinator process and worker processes?

specs/lifespan.rst
```diff
@@ -0,0 +1,55 @@
...
+
+Scope
+'''''
+
+The lifespan scope exists for the duration of the event loop. The
+scope itself contains basic metadata,
+
+* ``type``: ``lifespan``
+
+* ``asgi``: A dictionary containing a ``version`` key, which itself is
```
andrewgodwin: Don't think this needs to be redefined in every protocol.

specs/lifespan.rst
```diff
@@ -0,0 +1,55 @@
+=================
+Lifespan Protocol
+=================
+
+**Version**: 2.1 (2018-07-01)
```
andrewgodwin: I'd give this its own version. The point of protocols is to allow separate versioning rather than moving them all in lockstep.

specs/lifespan.rst
```diff
@@ -0,0 +1,55 @@
...
+* ``type``: ``lifespan``
+
+* ``asgi``: A dictionary containing a ``version`` key, which itself is
+  a unicode string version identifier, defaults to ``2.0``.
+
+
+Startup
+'''''''
+
+Sent when the server is ready to startup and recieve connections, but
```
andrewgodwin: What would this mean if the server didn't have incoming connections, but instead called out to an external API (like some of the chatbot servers I've whipped up)? Is there a better way to phrase "connections"?

specs/asgi.rst
pgjones: Changed.

specs/lifespan.rst
pgjones: I've tried making this clearer.

specs/lifespan.rst
pgjones: Changed.

specs/lifespan.rst
pgjones: Done.

specs/lifespan.rst
pgjones: I think we should also change the asgi file as well, I was matching 

> This first callable is called whenever a new connection comes in to the protocol server, and creates a new instance of the application per connection (the instance is the object that this first callable returns).

I'm not sure how else to word it though, maybe just in terms of application instances?

tomchristie:
So thoughts came up, might be worth talking through...

1. As currently proposed there’s no “run some initialisation, and only start serving requests once that’s complete”. Do we need to provide for that, and if so, how? Could we mandate that the ASGI instatiation for this scope type is run to completion prior to any other ASGI messaging occurring?

2. Do we need a “start” message at all? Calling the application is the start. Perhaps we only need a shutdown?

3. Is referring to the lifespan as being in the context of the asyncio loop overly specific. Could we just slightly underspecify it, but indicate that it bounds the server lifetime, and that no other ASGI messages will be sent either side of the lifetime?

4. Would ~1 second heartbeat messages be useful in this context? It might provide a nice style for running cron-like background tasks? Is this out of scope?

andrewgodwin:
Some thoughts on your thoughts, Tom:

1. I do think waiting for init to complete is sensible, but there's the problem that if you want to wait on an ASGI application for something you have to have it send you a message back (you can't block on the message you sent to it). That may be negated by 2, though.

2. In all the other ASGI applications I've tried to avoid doing something in that initial call as it's in a sync context, not an async context, and so you can't do non-blocking things there easily. This is, however, less of a problem for server start if we say it must be blocking.

3. I think that might be preferable.

4. Maybe a heartbeat is out of scope, but I can definitely see the utility. Of course, picking the interval is the trick - 1 second doesn't seem like a bad choice, though, but we'd have to make sure people didn't write apps that did stuff that took more than a second on the heartbeat and got backed up (so that when they quit the server, the app is still getting through its heartbeat backlog and can't run the shutdown code in time). For that reason, it's conceivable the heartbeat functionality should be separate.

tomchristie:
Agreed (1-3), and yes would consider (4) heartbeat as out of scope.

andrewgodwin:
OK - @pgjones, do you agree with all that and want to change the PR to match?

pgjones:
Sorry, I've not caught up yet, I should be able to look next week (at EuroPython this week).

pgjones:
Having re-read these I think that #45 might be the better solution. The problem I see is with case 2), as I want an explicitly async startup context in the event loop the server will use. I'm not sure we need a sync startup context as any call and a lock would work (like the Flask before first request system). 

Alternatively we could add to this that the app should respond with say `lifespan.startup.complete` and `lifespan.cleanup.complete` with the server waiting. #45 is simpler and clearer in my view though.

tomchristie:
Sending ‘startup’ and ‘shutdown’ messages would be sufficient for applications to await a complete startup before sending other message types.

pgjones:
I don't follow you, how would the application await them without the serving sending other message types?

tomchristie:
I mean this *sort* of pattern is possible with a ‘lifecycle’ scope and a single ‘shutdown’ message.

```
class LifecycleApp:
    started = asyncio.Event()  # class attribute

    def __init__(self, scope):
        self.scope = scope

        if self.scope[‘type’] != ‘lifecycle’:
            await self.started.wait()
            await self.main(receive, send)  # Not implemented here
        else:
             await self.startup(). # Not implemented here.
             self.started.set()
             message = await receive()
             assert message[‘type’] == ‘shutdown’
             await self.shutdown(). # Not implemented here.
```

pgjones:
I've split this into #60 and #61. I think #61 has all the features discussed here, but slightly miss-uses the ASGI double callable.

PR #53, opened 2018-07-14
Allow for http.disconnect after response sending

pgjones:
This allows for ASGI applications to call the receive after sending a
response by indicating the application should receive a disconnect
message. This in turn allows the sending and receiving tasks to have
separate lifespans.

The use case I have in mind (and use in Quart) is something like,
```
async def application_instance(receive, send):
    while True:
        message = await receive()
        task = asyncio.ensure_future(sending_task(send))  # Separate lifespan
        if message['type'] == 'http.disconnect':
            task.cancel()
            break
        ...
```

At the moment Uvicorn (as the motivating example) interprets this a different want and raises an error in the receive after the sending is complete.

### DIFF
--- a/specs/www.rst
+++ b/specs/www.rst
@@ -152,9 +152,10 @@ Keys:
 Disconnect
 ''''''''''
 
-Sent to the application when a HTTP connection is closed. This is mainly
-useful for long-polling, where you may want to trigger cleanup code if
-the connection closes early.
+Sent to the application when a HTTP connection is closed or if receive
+is called after a response has been sent. This is mainly useful for
+long-polling, where you may want to trigger cleanup code if the
+connection closes early.
 
 Keys:
 


tomchristie:
Out of interest: What’s the motivation for knowing if the client has disconnected here? In the context of running a post-request background task it’s not obvious why it’d be relevant. What am I missing?

andrewgodwin:
Are you just seeing this as clarifying that the only valid message to receive after you've sent a response would be a disconnect? I might reword it slightly to make it clear that calling receive after the response is sent does not necessarily have to _immediately_ return a disconnect, just that it's the only valid message that might occur at that point.

pgjones:
I mostly want to clarify the two cases in https://github.com/encode/uvicorn/pull/126. To do so I think the application should always receive the disconnect message and that the `receive` call should block until it has a new message to send. I'm struggling to figure out where to put this in the specification though.

I think the application should always receive the disconnect message as it makes it much easier to figure out when to stop the asgi_instance. Otherwise, as with the snippet above, the application instance would have to check `self.task` for completeness and for disconnect messages from the `receive`. I agree with @tomchristie's point that it otherwise has no obvious use, but think it has no obvious downsides either.

I think the `receive` call should block as a client can get into a sequence that never yields to the event loop, as with the snippet above, if the server keeps sending the same message.

Note it is probably easier to understand my thoughts if you assume the `sending_task` function will send an infinite stream e.g. Server Sent Events or a websocket that only the client can close.

tomchristie:
If an application has not consumed the request body and calls `recieve` after a response has been sent, there are a few different things that *could* happen.

1. The server continues to stream the response body to completion, then blocks until the underlying connection is closed, at which point `http.disconnect` occurs.
2. The server continues to stream the response body to completion, then responds with `http.disconnect` (Not reflecting the underlying connection, but reflecting the request/response cycle.)
3. The server responds with `http.disconnect` immediately.

The last one has the benefit of servers no longer buffering request bodies once an application has sent its response. (Particularly useful for eg. not buffering large request bodies in cases where the application does not consume the request body.)

Currently `uvicorn` actually does:

4. Raise an error to the application, since I’d (incorrectly) assumed that calling `receive` after having sent the response would necessarily indicate an error.

My inclination is that either (2) or (3) should both be valid behaviours, and that `http.disconnect` should probably represent *either* the end of the request/response cycle *or* any early client disconnect. (Whichever comes first)

(Although I’m not clear if the intent is that it strictly reflects the condition of the underlying connection)

andrewgodwin:
Right, I agree with your inclination, Tom.

pgjones:
Reading through the comments, I think this change as worded matches the inclination. I think it also helps clarify when `http.disconnect` can be received. Is it good to merge?

andrewgodwin:
Yes, I think it's good to go - merged!

PR #60, opened 2018-08-20
Add ASGI versioning to the scope definition

pgjones:
This allows frameworks to know what version of the ASGI specification
the server implements. This in turn allows features to be added to the
specification for framework usage.


andrewgodwin:
Can we remove the "as this was added in version 2.1" text and just say it defaults to 2.0? I don't think this PR will cause a version bump.

pgjones:
Yep, done.

andrewgodwin:
Great, thanks!

PR #61, opened 2018-08-20
Add startup and cleanup procotols

pgjones:
This allows for frameworks to provide the ability to call async code
to setup before the first connection, e.g. creating a connection pool,
and cleanup after all connections have been closed, e.g. releasing the
connection pool.

### DIFF
--- a/specs/asgi.rst
+++ b/specs/asgi.rst
@@ -214,6 +214,7 @@ Current protocol specifications:
 
 * :doc:`HTTP and WebSocket <www>`
 
+* :doc:`Startup and Cleanup (Lifespan) <lifespan>`
 
 Middleware
 ----------
--- a/specs/lifespan.rst
+++ b/specs/lifespan.rst
@@ -1,0 +1,53 @@
+==============================
+Startup and Cleanup (Lifespan)
+==============================
+
+**Version**: 1.0 (2018-08-20)
+
+The startup and cleanup ASGI sub-specification outlines how to
+communicate the lifespan events startup and cleanup within ASGI. The
+lifespan being referred to is that of main event loop. In a
+multi-process environment there will be lifespan events in each
+process.
+
+These sub-specifications allow an application to complete startup or
+cleanup whilst the server waits. They differ from the WWW
+sub-specifications in that no events are sent, with the
+application_instance callable awaited by the server till completion.
+
+In order to be backwards compatible a server should continue if an
+application errors when receiving these scopes.
+
+Startup
+-------
+
+The startup sub-specification is simply a scope with the following,
+
+* ``type``: ``lifespan``
+
+the callable returned by the application will then be awaited by the
+server. On completion of the callable the server will start accepting
+connections.
+
+Cleanup
+-------
+
+The cleanup sub-specification is simply a scope with the following,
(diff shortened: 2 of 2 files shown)


andrewgodwin:
Did we get to the point of "awaiting the second callable" last time? I thought the idea was to either await the first callable or to send an event in - the second callable in all applications is going to want `send` and `receive` awaitables as part of its signature.

tomchristie:
> In order to be backwards compatible a server should continue if an application errors when receiving these scopes.

Something else to consider - errors raised during startup ought to typically result in a halt.

tomchristie:
Summarizing different possible approaches:

* `async startup()`, `async shutdown()` attributes on the ASGI callable itself. Nice and simple for class-based implementations, no backwards-compat issues, but doesn't really work if middleware is used.
* A single lifespan scope, with `startup`/`shutdown` events (tho first of those *could* just be implicit.) Really the server needs to be flagged on startup completion so you possible want a pair of messages to send back to indicate "startup_complete"/"shutdown_complete". (tho second of those *could* just be implicit.)
* A pair of `startup`/`shutdown` scopes, with no messaging used on receive/send channels. (Ie. this proposal.)

tomchristie:
Having slept on it, I'd personally favor (2) for a few reasons:

* Backwards compatible. Ignore it if an error is raised instantiating the ASGI app. Halt if an error occurs during the actual lifecycle.
* Mirrors the websocket style, fits in with other ASGI patterns, and event-driven style.
* Debugs nicely. A server at debug log level might start by displaying a pair of "startup"/"startup complete" messages, then a bunch of http/ws, then end with a pair of "shutdown"/"shutdown" complete.
* Works with middleware, which can pass the messages along the chain if needed.
* It's easy to implement an app pattern than awaits `startup`/`shutdown` methods on the events, eg...

```python
class App:
    def __init__(self, scope):
        self.scope = scope

        if self.scope['type'] == 'lifecycle':
            while True:
                 message = await receive()
                 if message['type'] == 'lifecycle.startup':
                     await self.startup()
                     await send({'type': 'lifecycle.startup.complete'})
                 elif message['type'] == 'lifecycle.shutdown':
                     await self.shutdown()
                     await send({'type': 'lifecycle.shutdown.complete'})
                     return
        else:
            # Handle other types

    async def startup(self):
        ...

    async def shutdown(self):
        ...
```

It's a more verbose approach than the other two options, but I think it's more in keeping with the rest of the ASGI spec.

Thoughts @pgjones, @andrewgodwin?

andrewgodwin:
I also think I prefer a single lifecycle protocol with events, and having the protocol send back events when things can continue. That way, the startup code is also free to just say "continue" as the first thing and do its stuff in the background as well, if it doesn't need blocking behaviour.

pgjones:
I like this approach as well, what has put me off so far is that it isn't clear how existing applications should interact with servers that introduce this protocol. I think existing ASGI clients will throw some kind of error, so I suggest the server just log the error and carry on (not block waiting for the `lifecycle.startup.complete` message). If this seems good, I'll write it up.

pgjones:
Ok, so hopefully we are there with #63.

PR #62, opened 2018-08-24
Run tests against Python 3.7 and add classifier

michael-k:


andrewgodwin:
Why not add "3.7" to the Python versions set further up the file instead?

michael-k:
Because travis does not support this (yet); see travis-ci/travis-ci#9815
It's only possible to add "3.7-dev" to the list of Python versions directly, but that's a very old version and all maintainers I've interacted with, preferred the workaround over 3.7-dev

andrewgodwin:
OK, thanks

PR #63, opened 2018-09-06
Add a lifespan protocol

pgjones:
This allows for frameworks to provide the ability to call async code
to setup before a single request, e.g. creating a connection pool, and
cleanup after all requests, e.g. releasing the connection pool
connections.

### DIFF
--- a/specs/lifespan.rst
+++ b/specs/lifespan.rst
@@ -1,0 +1,112 @@
+=================
+Lifespan Protocol
+=================
+
+**Version**: 1.0 (2018-09-06)
+
+The Lifespan ASGI sub-specification outlines how to communicate
+lifespan events such as startup and cleanup within ASGI. The lifespan
+being referred to is that of main event loop. In a multi-process
+environment there will be lifespan events in each process.
+
+The lifespan messages allow for a application to initialise and
+cleanup in the context of a running event loop. An example of this
+would be creating a connection pool and subsequently closing the
+connection pool to release the connections.
+
+A possible implementation of this protocol is given below::
+
+    class App:
+
+        def __init__(self, scope):
+            self.scope = scope
+
+        async def __call__(self, receive, send):
+            if self.scope['type'] == 'lifespan':
+                while True:
+                    message = await receive()
+                    if message['type'] == 'lifespan.startup':
+                        await self.startup()
+                        await send({'type': 'lifespan.startup.complete'})
+                    elif message['type'] == 'lifespan.cleanup':
+                        await self.cleanup()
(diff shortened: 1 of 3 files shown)


specs/lifespan.rst
```diff
@@ -0,0 +1,111 @@
...
+
+Scope
+'''''
+
+The lifespan scope exists for the duration of the event loop. The
+scope itself contains basic metadata,
+
+* ``type``: ``lifespan``
+
+Servers must ignore any exceptions raised when calling the application
```
andrewgodwin: I'd add a comment further saying that if an application errors with the lifespan protocol then the server should explicitly continue but not run lifespan events, to make it crystal clear.

pgjones:
I've made the clarifying change.

andrewgodwin:
:+1:

tomchristie:
Should I expect this to be showing on readthedocs? Doesn't seem to be up there...

https://asgi.readthedocs.io/en/latest/extensions.html

tomchristie:
For context: https://www.starlette.io/events/

Having implemented this into uvicorn and starlette, couple of observations.

* Should we prefer `startup`/`shutdown` over `startup`/`cleanup`?
* The "if you get an exception on application init" is sufficient, but a bit loose. Eg. Sometimes in middleware you'll end up with this sort of pattern...

```python
class ExceptionLoggerMiddleware:
    def __init__(self, app):
        self.app = app

    def __call__(self, scope):
        return functools.partial(self.run_asgi, scope=scope)

    async def run_asgi(self, receive, send, scope=None):
        try:
            inner = self.app(scope)
            await inner(receive, send)
        except:
            ...
```

If you have something like that between the server and the application, *and* if the application doesn't support the lifecycle protocol, than the server will see the lifecycle init return okay, but then error when it is called, indicating a startup error rather than a "does not support lifecycle protocol".

Not a showstopper by any means, but worth being aware of.

tomchristie:
Sounds good. Slight preference from me for startup/shutdown, but happy with whatever Andrew goes for. Look forward to seeing it in Hypercorn. I might have a dig into getting together a PR for daphne.

tomchristie:
Also possible: 'setup'/'teardown'

andrewgodwin:
I'd probably prefer startup/shutdown - cleanup could be some periodic cleanup thing as well. If we all agree then I will change it to be `shutdown`.

It should also appear on the docs now as I added the missing spec-include file that makes it work inside docs in dd288e0

tomchristie:
@andrewgodwin Works for me. Thanks.

andrewgodwin:
OK, it has been changed to `startup`/`shutdown`

PR #67, opened 2018-09-26
Drop API Star from implementations.

tomchristie:
Drop API Star, since from 0.6 onwards it's [descoped into a framework-agnostic suite of API tools](https://github.com/encode/apistar/pull/624), rather than a complete server implementation. (Starlette has picked up the mantle for my own work on an ASGI framework implementation.)


andrewgodwin:
👍

PR #69, opened 2018-10-24
Fix cleanup/shutdown

almarklein:
I just found out that I've implemented the lifespan protocol the wrong way in [Asgish](https://github.com/almarklein/asgish), because I misread the docs. This change fixes some ambiguities with earlier naming.

### DIFF
--- a/specs/lifespan.rst
+++ b/specs/lifespan.rst
@@ -5,12 +5,12 @@ Lifespan Protocol
 **Version**: 1.0 (2018-09-06)
 
 The Lifespan ASGI sub-specification outlines how to communicate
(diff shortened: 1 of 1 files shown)


almarklein:
Actually, maybe I read it the wrong way because I looked at the guide before the rename :)

pgjones:
Looks good to me.

andrewgodwin:
Thanks - looks like I wasn't perfect about the rename, and sorry we did it as late as we did!

almarklein:
No problem! (I like "shutdown" more too :)) Thanks for all the work y'all are doing on ASGI!

Issue #71, opened 2018-11-07
ThreadPool execution should preserve contextvar context.


tomchristie:
This is an implementation pattern, rather than something strictly for the ASGI spec.

Something that cropped up while implementing https://github.com/encode/sentry-asgi is that Python's ThreadPoolExecutor does not default to preserving the contextvars context, which *really* is what you want.

If you're using something like the Sentry SDK, then you needs `contextvars` in order to properly tie otherwise unrelated events together into, and you need it to continue working properly when threadpools are used.

Here's my fix in Starlette: https://github.com/encode/starlette/pull/192

Other ASGI tools and frameworks will want to use something similar when running sync code in threadpools.

Links to core issue here: https://github.com/encode/starlette/issues/191

andrewgodwin:
Given that this is what `sync_to_async` does, are you suggesting we should apply it there? That's the only way Channels runs things in threadpools, I believe.

airhorns:
FWIW we got bitten by this in our django-channels app and we're glad to see it fixed in the framework such that `ContextVar`s just work as you'd expect them across that thread boundary. We used https://pypi.org/project/contextvars-executor/ and a 

```python
loop = asyncio.get_event_loop()
loop.set_default_executor(ContextVarExecutor(max_workers=int(os.environ.get("ASGI_THREADS", 1))))
```
to fix this for anything using asyncio internally, but I think this fix means we can get rid of that. +1 for Python changing this sooner rather than later!

PR #73, opened 2018-12-03
Threadpool operations should run in same contextvars context.

tomchristie:
Refs #71

In versions of Python that support contextvars, threadpool operations should run in the same the context as the parent.

Prompted by https://github.com/getsentry/sentry-python/issues/162#issuecomment-443415106

Note that Python 3.8 might include [a `preserve_context` keyword argument](https://bugs.python.org/issue34014).

### DIFF
--- a/asgiref/sync.py
+++ b/asgiref/sync.py
@@ -3,6 +3,10 @@ import functools
 import os
 import threading
 from concurrent.futures import Future, ThreadPoolExecutor
+try:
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Looks good to me, just needs a little finessing to pass the linting!

andrewgodwin:
Thanks for doing this for me :)

PR #74, opened 2018-12-22
Clarify www spec for path definition

pgjones:
RFC 7320 section 5.3 allows for the request target to be in an
absolute form (5.3.1) which servers MUST allow. The current ASGI
servers (Hypercorn, Uvicorn & Daphne) will all pass on the full
request-target (minus query string) to the ASGI application rather
than, as the previous version of this specification suggested, just
the path part of the URL. This should help ASGI app developers
understand whether to assume the path will always be just the path.


andrewgodwin:
Lovely! Thanks.

PR #76, opened 2019-01-17
Add http_version to the Websocket scope

pgjones:
This is required now that websockets over HTTP/2 has been defined in
RFC 8441.

### DIFF
--- a/specs/www.rst
+++ b/specs/www.rst
@@ -189,6 +189,9 @@ contains the initial connection metadata (mostly from the HTTP handshake):
 
 * ``type``: ``websocket``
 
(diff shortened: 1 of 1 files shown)


andrewgodwin:
We can't add a new key without either bumping the spec version or giving it a default. Want to add "optional, default is 1.1"?

pgjones:
Yep, done.

andrewgodwin:
Lovely, thanks!

ryomahan:
Do we need to add a reminder after this sentence to explain why we need a "http_version" key in websocket scope.

PR #79, opened 2019-02-22
Allow headers to be specified in the accept message

pgjones:
This allows an application to specify additional headers to send to
the client in the handshake response, for example cookie settings.

(See https://gitlab.com/pgjones/quart/issues/206 for a motivating request).

### DIFF
--- a/specs/www.rst
+++ b/specs/www.rst
@@ -252,6 +252,13 @@ Sent by the application when it wishes to accept an incoming connection.
 * ``subprotocol``: The subprotocol the server wishes to accept, as a unicode
   string. Optional, defaults to ``None``.
 
+* ``headers``: A list of ``[name, value]`` lists, where ``name`` is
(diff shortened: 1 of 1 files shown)


pgjones:
I'm fairly confident the test failures are unrelated.

andrewgodwin:
I agree with this, but it does feel like we're going to need a version sent with the protocols soon (as well as the one with the overall ASGI handshake) - otherwise apps are going to try and send this into the void (or worse, into a server exception).

I've fixed the tests, so if you rebase that should go green.

pgjones:
Maybe this should be bundled with the version 3 change?

andrewgodwin:
We can definitely do that, but I'm not sure I want to ratchet top-level ASGI specs numbers to get HTTP/WebSocket improvements. I'll merge this in and then add a version number to the `scope`.

PR #80, opened 2019-03-04
3.0: Single-callable style applications & lifecycle improvements

andrewgodwin:
This branch both adds the doc/spec change for single-callable-style applications, as well as upgrading the lifecycle protocol to have better failure modes.

### DIFF
--- a/specs/asgi.rst
+++ b/specs/asgi.rst
@@ -233,7 +233,7 @@ defined by the protocol spec. Examples of a message ``type`` value include
 
   Applications should actively reject any protocol that they do not understand
   with an Exception (of any type). Failure to do this may result in the server
-  thinking you support a protocol you don't, which is especially dangerous with
+  thinking you support a protocol you don't, which can be confusing with
   the Lifespan protocol, as the server will wait to start until you tell it.
 
 
--- a/specs/asgi.rst
+++ b/specs/asgi.rst
@@ -229,6 +229,14 @@ where the ``protocol`` matches the scope type, and ``message_type`` is
 defined by the protocol spec. Examples of a message ``type`` value include
 ``http.request`` and ``websocket.send``.
 
+.. note::
+
+  Applications should actively reject any protocol that they do not understand
+  with an Exception (of any type). Failure to do this may result in the server
+  thinking you support a protocol you don't, which is especially dangerous with
+  the Lifespan protocol, as the server will wait to start until you tell it.
+
+
 Current protocol specifications:
 
 * :doc:`HTTP and WebSocket <www>`
@@ -356,6 +364,7 @@ unicode strings.
 Version History
 ===============
 
+* 3.0 (2019-03-04): Changed to single-callable application style
 * 2.0 (2017-11-28): Initial non-channel-layer based ASGI spec
 
 
--- a/specs/asgi.rst
+++ b/specs/asgi.rst
@@ -2,7 +2,7 @@
 ASGI (Asynchronous Server Gateway Interface) Specification
 ==========================================================
 
-**Version**: 2.0 (2017-11-28)
+**Version**: 3.0 (2019-03-04)
 
 Abstract
 ========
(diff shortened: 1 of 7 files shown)


specs/lifespan.rst
```diff
@@ -78,6 +82,18 @@ Keys:
...
+Startup Failed
+''''''''''''''
+
+Sent by the application when it has failed to complete its startup. If a server
+sees this it should log/print the message provided and then exit.
+
+Keys:
+
+* ``type``: ``lifespan.startup.failed``
+* ``message``: Unicode string (optional, defaults to ``""``).
```
tomchristie: Wondering if we should use `"message": str` or `"exception": Exception` here:

* It might make sense to defer to the servers' exception logging.
* Passing through the exception would mean that we'd be able to give exception middleware (eg. Sentry) visibility onto startup/shutdown errors.

?

specs/asgi.rst
```diff
@@ -217,6 +229,14 @@ where the ``protocol`` matches the scope type, and ``message_type`` is
 defined by the protocol spec. Examples of a message ``type`` value include
 ``http.request`` and ``websocket.send``.
 
+.. note::
+
+  Applications should actively reject any protocol that they do not understand
+  with an Exception (of any type). Failure to do this may result in the server
+  thinking you support a protocol you don't, which is especially dangerous with
+  the Lifespan protocol, as the server will wait to start until you tell it.
```
tomchristie: Wondering if there might be an alternate to: "which is especially dangerous with the Lifespan protocol, as the server will wait to start until you tell it." - I think wording it as a *dangerous* case is maybe overly strong?

specs/lifespan.rst
tomchristie: Possibly related to this, possibly not... We've managed so far without an explicit "Here's an exception" message, and instead just raise the exceptions, which works great so far. But I think ASGI will eventually want one, because:

* If a client runs any concurrent-branching code (any protocol type) it may need to communicate 0, 1, or many exceptions up the stack.
* If the client runs any timer-driven or clock-driven events within the context of the `lifespan` task, it may need to communicate exceptions up the stack, without actually raising an exception and terminating the task.

(I don't think an exception message type *neccessarily* needs to be a 3.0 thing, but may be relevant to considering which of `message` vs `exc` is a sensible choice here.)

tomchristie: Perhaps the synposis to that oughta be: 1. "Let's use `message` for startup/shutdown failures" and 2. "Let's *plan* on introducing an exception message type (valid on all protocols)" ?

andrewgodwin: The ASGI spec says you're not allowed to send anything outside of basic types (strings, ints, etc.) in messages. We'd need to extend that a lot to allow arbitary Exception classes. I think Message should be enough for now?

And yes, I agree we probably want a generic "this is an exception" message at some point.

specs/asgi.rst
andrewgodwin: True, it's not dangerous per se. I'll reword to say "confusing".

specs/lifespan.rst
tomchristie: 👍

specs/lifespan.rst
```diff
@@ -46,14 +46,18 @@ Scope
 '''''
 
 The lifespan scope exists for the duration of the event loop. The
-scope itself contains basic metadata,
+scope itself contains basic metadata:
 
 * ``type``: ``lifespan``
 
 If an exception is raised when calling the application callable with a
-lifespan scope the server must continue but not send any lifespan
```
pgjones: I'm not sure this makes sense - how would you call the application with a ``lifespan.startup`` message?
Do I understand that any exceptions raised by an application handling the lifespan protocol should be ignored, with only the failed message counting?

specs/asgi.rst
```diff
@@ -143,46 +143,29 @@ serializable, and so they are only allowed to contain the following types:
...
+.. note::
 
-    application(scope)
-
-* ``scope``: The Connection Scope, a dictionary that contains at least a
-  ``type`` key specifying the protocol that is incoming.
+    The application format changed in 3.0 to use a single callable, rather than
+    the prior two-callable format. Two-callable is documented below in
+    "Legacy Applications"; servers can easily implement support for it using
+    the ``asgiref.compatibility`` library, and should try to support it.
```
pgjones: I'm not sure we should try to support both, I'm thinking it would be a pain specifically in middleware to do so. I'm tempted to take advantage of some breaking changes in Hypercorn to switch to asgi 3 only.

tomchristie: >  I'm thinking it would be a pain specifically in middleware to do so.

No doubt. I think the note's more intended at server implementations. I don't think it's a particularly big deal if Hypercorn prefers to go 3.0 only, so long as there's alternatives providing a graceful 2-or-3 policy. (I'll plan to go 2-or-3 with Uvicorn)

specs/lifespan.rst
tomchristie: A server could reasonably either:

* Ignore the exception, log that it was disabling lifespan, and carry on as normal.
* Exit and inform the developer that they need to run disabling the lifespan protocol.

The important thing is that applications have an explicit way of communicating "The server attempted to startup and failed", which needs to be distinct from "I don't understand/support lifespan, I'd better raise an exception."

What's important here is that simple HTTP-only clients should be able to function gracefully, without the developer neccessarily having remembered to disable the lifespan protocol.

specs/asgi.rst
pgjones: Actually, I'm happy with this wording. I think I'll support both in Hypercorn.

pgjones:
Should this also mention/empahsise that the server is responsible for setting the asgi version in the scope?

specs/lifespan.rst
```diff
@@ -2,7 +2,7 @@
 Lifespan Protocol
 =================
 
-**Version**: 1.0 (2018-09-06)
+**Version**: 2.0 (2019-03-04)
```
pgjones: @andrewgodwin Just a note/reminder we need to add the `spec_version` to the scope.

pgjones:
This is the [Hypercorn](https://gitlab.com/pgjones/hypercorn/commit/d47b2033e7db0eacbd7118534a35ef7dbe3bf694) implementation, for reference.

andrewgodwin:
@pgjones Just added spec_version and version everywhere I could, to reinforce that it's needed. I think that will be sufficient until I can write a proper compliance-testing app.

If there's no more feedback from you or @tomchristie, then I think it's time to merge this and anoint ASGI 3.0.

tomchristie:
I've taken a first crack at getting this into Daphne - https://github.com/django/daphne/pull/255
I can follow up on that if helpful, tho I'm a bit uncertain about how to go about adding the tests for it.

specs/asgi.rst
```diff
@@ -194,11 +177,40 @@ the server implements. If missing the version should default to
 ``"2.0"``.
```
tomchristie: I guess this needs to change?

PR #83, opened 2019-04-20
Clarify general and spec documentation

daneah:
Hey folks, I'm keen on making myself useful toward getting ASGI into Django (I talked with @andrewgodwin at DjangoCon US 2018 about it and recently saw @tomchristie's DjangoCon EU 2019 keynote) and I'm starting pretty well from scratch by reading the ASGI specification docs. Since I've just gone through the thing, I wanted to contribute some clarifying changes to things that tripped me up that I hope will provide some value to others.

The biggest change was moving data types for protocol specifications to a consistent place for better readability, e.g.:

### Before

* `type`: `http`
* `asgi["version"]`: The version of the ASGI spec, as a string

### After

* `type` (_Unicode string_) – `"http"`
* `asgi["version"]` (_Unicode string_) – Version of the ASGI spec

The rest being mostly minor stuff for consistency:

* Fix a few typographical errors, punctuation, etc.
* Rephrase a few passages for clarity
* Use consistent capitalization of "Unicode"
* Use consistent phrasing for "percent encoding"
* Change double-negative "not decoded"/"not unencoded" to "encoded"

Hope this helps; I'll be reading more on Starlette, uvicorn, etc. as well as Django roadmap stuff where I can, and if there are any specific areas you need help with presently please don't hesitate to let me know. Cheers!

### DIFF
--- a/docs/introduction.rst
+++ b/docs/introduction.rst
@@ -29,7 +29,7 @@ receiving WebSocket frames) can't trigger this.
 How does ASGI work?
 -------------------
 
-ASGI is structured as a double-callable---the first callable takes a
+ASGI is structured as a double-callable - the first callable takes a
 ``scope``, which contains details about the incoming request, and returns a
 second, coroutine callable. This second callable is given ``send`` and
 ``receive`` awaitables that allow the coroutine to monitor incoming events and
--- a/docs/introduction.rst
+++ b/docs/introduction.rst
@@ -29,11 +29,11 @@ receiving WebSocket frames) can't trigger this.
 How does ASGI work?
 -------------------
 
-ASGI is structured as a double-callable - the first callable takes a ``scope``,
-which contains details about the incoming request, and returns a second,
-coroutine callable. This second callable gets given ``send`` and ``receive``
-awaitables that allow the coroutine to monitor incoming events and send
-outgoing events.
+ASGI is structured as a double-callable---the first callable takes a
+``scope``, which contains details about the incoming request, and returns a
+second, coroutine callable. This second callable is given ``send`` and
+``receive`` awaitables that allow the coroutine to monitor incoming events and
+send outgoing events.
 
 This not only allows multiple incoming events and outgoing events for each
 application, but also allows for a background coroutine so the application can
@@ -56,9 +56,9 @@ Every *event* that you send or receive is a Python ``dict``, with a predefined
 format. It's these event formats that form the basis of the standard, and allow
 applications to be swappable between servers.
 
-These *events* each have a defined ``type`` key, so you know what it is to look
-for. Here's an example event for sending the start of a HTTP response that
-you might get from ``receive``::
+These *events* each have a defined ``type`` key, which can be used to infer
+the event's structure. Here's an example event for sending the start of a HTTP
+response that you might get from ``receive``::
 
     {
         "type": "http.response.start",
--- a/specs/asgi.rst
+++ b/specs/asgi.rst
@@ -235,8 +235,9 @@ defined by the protocol spec. Examples of a message ``type`` value include
 
   Applications should actively reject any protocol that they do not understand
   with an `Exception` (of any type). Failure to do this may result in the
-  server thinking you support a protocol you don't, which can complicate the
-  Lifespan protocol, as the server will wait to start until you tell it.
+  server thinking you support a protocol you don't, which can be confusing when
+  using with the Lifespan protocol, as the server will wait to start until you
+  tell it.
 
 
 Current protocol specifications:
--- a/specs/asgi.rst
+++ b/specs/asgi.rst
@@ -114,14 +114,14 @@ Events
 ------
 
 ASGI decomposes protocols into a series of *events* that an application must
-react to. For HTTP, this is as simple as two events in order---``http.request``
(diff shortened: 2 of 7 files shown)


daneah:
Incidentally I had to manually install `sphinx` and comment out some stuff in `docs/conf.py` to get the docs building locally; is there a pre-baked something that I'm missing around that?

docs/introduction.rst
```diff
@@ -29,11 +29,11 @@ receiving WebSocket frames) can't trigger this.
 How does ASGI work?
 -------------------
 
-ASGI is structured as a double-callable - the first callable takes a ``scope``,
-which contains details about the incoming request, and returns a second,
-coroutine callable. This second callable gets given ``send`` and ``receive``
-awaitables that allow the coroutine to monitor incoming events and send
-outgoing events.
+ASGI is structured as a double-callable---the first callable takes a
```
andrewgodwin: Not a huge fan of en or em dashes, they have confusing subtlety for those unfamiliar with them. Let's keep the normal dashes, and use semicolons or periods if a bigger context break is needed?

docs/introduction.rst
```diff
@@ -29,11 +29,11 @@ receiving WebSocket frames) can't trigger this.
...
+``scope``, which contains details about the incoming request, and returns a
+second, coroutine callable. This second callable is given ``send`` and
+``receive`` awaitables that allow the coroutine to monitor incoming events and
+send outgoing events.
```
andrewgodwin: I've just realised I need to revise this part of the intro for the new spec. I'll do that once this lands.

specs/asgi.rst
```diff
@@ -232,9 +232,9 @@ defined by the protocol spec. Examples of a message ``type`` value include
 .. note::
 
   Applications should actively reject any protocol that they do not understand
-  with an Exception (of any type). Failure to do this may result in the server
-  thinking you support a protocol you don't, which can be confusing with
-  the Lifespan protocol, as the server will wait to start until you tell it.
+  with an exception (of any type). Failure to do this may result in the server
```
andrewgodwin: `Exception` is actually the right capitalisation here as it's the name of the root Python type that is expected.

specs/asgi.rst
```diff
@@ -232,9 +232,9 @@ defined by the protocol spec. Examples of a message ``type`` value include
...
+  thinking you support a protocol you don't, which can complicate the Lifespan
```
andrewgodwin: I don't like complicate as a substitute for confusing; try rewording it? Needs to convey that this is not apparent to the end-user, and potentially dangerous.

specs/asgi.rst
```diff
@@ -306,26 +306,26 @@ There are times when protocol servers may want to provide server-specific
...
-only to indicate that the server accepts additional events via the ``send``
-callable, it may just be an empty dict.
+``extensions`` dictionary under the extension's name, and the value of that
+entry should itself be a ``dict``. Servers can provide any extra scope
+information that is part of the extension inside this value or, if the
+extension is only to indicate that the server accepts additional events via
+the ``send`` callable, it may just be an empty ``dict``.
 
-As an example, imagine a HTTP protocol server wishes to provide an extension
+As an example, imagine an HTTP protocol server wishes to provide an extension
```
andrewgodwin: I would not choose to put `an` before a strong `H` sound, like in HTTP.

docs/introduction.rst
daneah: Sounds good to me; this specific line is what put me on that path since "double-callable - the first" was visually busy. I'll back most of these dash changes out and replace with sentence breaks where it makes sense!

specs/asgi.rst
daneah: I had flipped back and forth on this one; perhaps stylizing as `Exception` vs. Exception is a good compromise there 👍

specs/asgi.rst
daneah: Makes sense! I think I might just back this change out given that additional context.

specs/asgi.rst
andrewgodwin: I agree with that.

specs/asgi.rst
daneah: Ah, this may be a British/American English difference I wasn't familiar with 😉I tried to avoid those on the whole, happy to back this out!

andrewgodwin: Yes, it's probably the *aitch* versus *haitch* pronunciation thing. While Django doc standard is US English, I would like to leave it as `a`, just for me :)

daneah:
Thanks for the feedback @andrewgodwin; those points should all be addressed! The error I was seeing after installing `sphinx` and running `make html` in `docs/` is the following:

```
...
Theme error:
An error happened in rendering the page extensions.
Reason: TemplateNotFound()
make: *** [html] Error 2
```

That led me [here](https://stackoverflow.com/a/45580325/2077393), which worked.

andrewgodwin:
Thanks! I'll merge it in.

As for the theme, I'm not sure why I would have that theme and a fresh install wouldn't. Maybe it's a package version thing?

daneah:
Yeah, possible! If I'm tinkering later I'll see if I can suss out what might be implicated there.

PR #85, opened 2019-04-25
Fix and document docs process

### LINKED ISSUE #84: Building the documentation
In #83 I mentioned that I needed to comment out a few lines in `docs/conf.py` to get them to build. `conf.py` currently specifies both `html_theme = 'default'` and `html_sidebars = {...}`, which I believe to be mutually exclusive. The simplest thing that would fix the build is commenting out `html_sidebars`, resulting in the default sphinx theme; alternatively, setting `html_theme` to `'sphinx_rtd_theme'` and documenting the `sphinx_rtd_theme` package as a docs dependency would produce docs that look like those on https://asgi.readthedocs.io.

Earlier versions of sphinx shipped with `sphinx_rtd_theme`, but it now ships with `alabaster`. Setting the theme to `sphinx_rtd_theme` and trying to build without the dependency states:

```
sphinx_rtd_theme is no longer a hard dependency since version 1.4.0. Please install it manually.(pip install sphinx_rtd_theme) 
```

I can open a pull request with either solution along with a mention of the docs process in the README. Wanted to open this first to see which is preferable!

daneah:
Resolves #84 

* Comment out `html_sidebars` lines from `docs/conf.py` to get default working Sphinx theme when building docs out of the box
* Add section in `README.rst` about how to build the docs (this can go somewhere else if it's starting to feel lengthy?)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -97,11 +97,36 @@ Contributing
 Please refer to the
 `main Channels contributing docs <https://github.com/django/channels/blob/master/CONTRIBUTING.rst>`_.
 
+
+Testing
+'''''''
+
 To run tests, make sure you have installed the ``tests`` extra with the package::
 
     cd asgiref/
     pip install -e .[tests]
     pytest
+
+
+Building the documentation
+'''''''''''''''''''''''''
+
+The documentation uses `Sphinx <http://www.sphinx-doc.org>`_::
+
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Thanks - looks good to me!

PR #89, opened 2019-05-06
ci: run linting after tests, output diff

blueyed:
It is not very useful to get a failure because of style issues, without
any feedback on the actual tests.  Also black should display what it
would change (`--diff`).


andrewgodwin:
Thanks!

PR #90, opened 2019-05-07
pytest: use testpaths instead of addopts

blueyed:
Nobody is using test selection?!


andrewgodwin:
I mostly use match-by-name, which worked fine, but thanks!

blueyed:
@andrewgodwin 
You mean `-k`?

Issue #87, opened 2019-05-01
Provide raw path / URI


blueyed:
The `path` currently gets unquoted / percent-decoded, which results in unrecoverable loss of information: a request to `/foo/bar` is the same as `/foo%2Fbar`.

With WSGI there seems to exist workarounds via `RAW_URI` / `REQUEST_URI` (e.g. in gunicorn, https://github.com/etianen/aiohttp-wsgi/pull/17).

I think it would be nice to have `raw_path` officially in the spec.
I am not sure if it should be a unicode string, or bytes (to avoid another field later on).

(I do not think that `raw_uri` (path + query info) is necessary, since `query_string` is already given as bytes)

Current spec: https://asgi.readthedocs.io/en/latest/specs/www.html#id1
(Via https://github.com/encode/uvicorn/pull/354)

andrewgodwin:
We already looked at this in https://github.com/django/asgiref/issues/51 - I am curious if there is a use for the raw path other than those discussed there. Specifically the fact that while WSGI claims to give you the raw path, this is not always actually true.

blueyed:
Oh, sorry for missing #51.  I've only skimmed this quickly - you said there:

> if we provided a "raw" one then people might rely on it only for it to break later on when a webserver in front did exactly the same thing (webservers: not good at adhering perfectly to the URI RFC)

So because others might break it it gets broken upfront? :)
It basically means that any proxy using ASGI is forced into the group of "webservers: not good at adhering perfectly to the URI RFC" then.

I can see that it is too late to change `path` now probably, but something like `raw_path` should be added.

I currently have to get it in via setting a custom header from nginx, updating the `request.url` from there then.  This works for me, but should not be necessary really.

andrewgodwin:
It's more that this falls into the bucket of "well, it's useful for 1% of people", and I'm not personally convinced the job of a spec is to get all those corner cases - e.g. ASGI also doesn't allow for certain kinds of responses that are valid in HTTP.

Is there an app you're developing that is not low-level HTTP but does need raw path? If I may ask, what specifically about the path do you need?

blueyed:
I think it is more like "99% do not notice / are not affected".
The app I've noticed this with is a proxy after all.. :)

blueyed:
[...]

PR #92, opened 2019-06-08
Optional raw_path, refs #87

simonw:


pgjones:
Could this be added to the WebSocket scope as well? Also should the spec_version be bumped?

andrewgodwin:
I don't see a need to change the spec_version as this is an optional extra field. Agreed about WebSockets, though. Also don't say that servers have to provide `None` if they don't support it; just say they don't have to provide it (no need to make them send None rather than just leaving out the key)

simonw:
I'm not too keen on saying "you can either set it to `None` or omit the key entirely" because, if I was implementing the spec, I'd want to know which of those two options I should pick. That's why I decided to go with documented `None`.

How about we specify that you either provide the `raw_path` key WITH a not-null value, or you omit the key entirely, but you're not allowed to return `"raw_path": None`?

simonw:
New copy (I copied it to the websocket section as well):

HTTP___WebSocket_ASGI_Message_Format_-_Online_reStructuredText_editor

davidism:
One somewhat confusing thing in at least Gunicorn is that while it will handle splitting `SCRIPT_NAME` and `PATH_INFO`, it will not modify `RAW_PATH`. So if the WSGI app decides to use it instead of the decoded path, it needs to do its own pre-processing otherwise routing won't work. Maybe we should specify a `RAW_SCRIPT_NAME` and `RAW_PATH_INFO` as well as the full URL?

andrewgodwin:
@simonw Most optional fields in the current ASGI spec are defined as I suggested - optional, and if they're missing here's the default value you use. I would like to continue that.

@davidism ASGI handles the script_name/path_info thing by just providing a `root_path` option (which is essentially `SCRIPT_NAME`); the `path` that is provided is already analagous to `PATH_INFO`. I defer to you if providing a `raw_root_path` is useful; I would really hope that people aren't mounting apps under paths that they can't trust the server's decoding of, but then I do generally have too much hope.

We should at least call out the fact that this change would make `raw_path` _substantially_ different from `path` in the case of an app being mounted somewhere.

davidism:
Ok, I misunderstood the `root_path` behavior, I don't think we need a raw version of that.

simonw:
OK, I'll update it to fit that convention.

simonw:
Banners_and_Alerts_and_HTTP___WebSocket_ASGI_Message_Format_-_Online_reStructuredText_editor

andrewgodwin:
OK, this looks good to me. Unless there are any other comments, I'll merge this in.

Issue #95, opened 2019-06-22
Who is responsible for implementing Chunked-Transfer-Encoding


stuxcrystal:
I read the ASGI-Spec and it was not clear who is responsible for implementing Chunked-Transfer-Encoding for the event `http.request`.

Is the framework ultimately responsible for parsing whatever is sent over the raw socket or is the server responsible for supporting Transfer-Encoding?

andrewgodwin:
Do you mean _inbound_ transfer encoding, or _outbound_? I presume you mean inbound, given you mentioned `request`.

My stance on this is that it's up to the framework to handle it, as the ASGI spec does not have high-level-enough messages in the HTTP extension to handle this - but it will give the framework the request body as it arrives, allowing it to be done there. This also reflects how WSGI handled it (which is to say, it didn't).

davidism:
WSGI servers ended up being the ones that handled decoding. In ASGI the distinction is clearer and it does make sense for the app to handle it. Would be nice if there was a sans-io library for this, maybe there is already.

tomchristie:
Uvicorn automatically uses `Transfer-Encoding: chunked` For streaming response with no Content-Length set. I’ve not double checked Daphne or Hypercorn’s behaviour. I’d generally assume(?) that it makes sense that way, since it’s a required level of baseline functionality for streaming responses, and there’s no real value in forcing that onto framework-level implementations.

stuxcrystal:
I think it'd be good if the ASGI-spec could explicitly define who is responsible for handling `Transfer-Encoding` otherwise, I think, nobody will implement it or worse: Both will handle it, mangling the message.

andrewgodwin:
Outbound/response chunked encoding should be handled by servers, as it is in WSGI - I believe this is what @tomchristie is referring to. For inbound chunked encoding, I think it should be applications, if they wish. I will make a PR that clarifies this.

andrewgodwin:
OK, so I have whipped up a PR for the current situation (https://github.com/django/asgiref/pull/96), but I'm not super thrilled about having things be so asymmetric - application-handled inbound, and server-handled outbound. Are there any advantages to doing it this way? Or should we have the servers handle the transfer-encoding overhead? (Especially as it's only for single hops, so there's no guarantee a reverse proxy isn't getting rid of it)

davidism:
Transfer encoding doesn't really strike me as an application-level concern, so if it's reasonable for the server to handle decoding and encoding, that seems preferable.

[...]

PR #96, opened 2019-06-23
Add notes about Transfer-Encoding: chunked

andrewgodwin:
This follows up from #95


tomchristie:
Worth noting that I think this’ll be out of sync with the actual current behaviour of uvicorn & hypercorn, since eg the h11 library deals with chunked encoding transparently in both directions. I’d need to double check the httptools behaviour (uvicorn has two alternative http implementations).

andrewgodwin:
@tomchristie Yeah, and honestly I think ASGI maybe _should_ say it's transparent in both directions. See my comments in #95.

PR #101, opened 2019-06-27
AsyncToSync: use asyncio._get_running_loop for check

blueyed:
It is available since Python 3.5, and makes the check much simpler
(especially not having to catch RuntimeError for the common case).

### DIFF
--- a/asgiref/sync.py
+++ b/asgiref/sync.py
@@ -38,16 +38,11 @@ class AsyncToSync:
 
     def __call__(self, *args, **kwargs):
         # You can't call AsyncToSync from a thread with a running event loop
-        try:
-            event_loop = asyncio.get_event_loop()
-        except RuntimeError:
-            pass
-        else:
-            if event_loop.is_running():
-                raise RuntimeError(
-                    "You cannot use AsyncToSync in the same thread as an async event loop - "
-                    "just await the async function directly."
-                )
+        if asyncio._get_running_loop() is not None:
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Thanks for catching that!

davidism:
Is it a good idea to use a "private" (`_`) undocumented method? It looks like 3.7 added a public `get_running_loop()` that would replace this entire block, but we can't depend on that for a long time.

blueyed:
@davidism 
It is not ideal, but quite OK - it is tested.
Also `get_running_loop` also might throw - just the other way around: https://github.com/python/cpython/blob/65e187239a6feb09840ed9c2ebee51ed151e690d/Lib/asyncio/events.py#L681-L690

andrewgodwin:
I'm not super thrilled about using the undocumented method, but it's likely still better than the code that I had that did it the same way anyway. Maybe the best compromise is to use `get_running_loop` if it's available, and fall back to this if not?

blueyed:
It has a docstring at least: https://github.com/python/cpython/blob/65e187239a6feb09840ed9c2ebee51ed151e690d/Lib/asyncio/events.py#L694 - and according to this is meant to be used by event loops (e.g. uvloop also uses it).

blueyed:
For reference: it was reverted in https://github.com/django/asgiref/commit/4a48f4fdb8bcc3e8bfa46aa122db97e9cf31eb65.

PR #102, opened 2019-07-03
Fix broken document `lifespan`

tkamenoko:
Lifespan's document was broken.


andrewgodwin:
Gosh, how did I screw up the rebase like that? Thanks for fixing it!

PR #106, opened 2019-07-12
Expose application instance on ApplicationCommunicator

richtier:


asgiref/wsgi.py
```diff
@@ -31,12 +31,14 @@ def __init__(self, wsgi_application):
...
     async def __call__(self, scope, receive, send):
         self.scope = scope
-        # Alright, wait for the http.request message
-        message = await receive()
-        if message["type"] != "http.request":
-            raise ValueError("WSGI wrapper received a non-HTTP-request message")
         # Wrap send so it can be called from the subthread
         self.sync_send = AsyncToSync(send)
+        message = await receive()
+        if message["type"] == 'give-me-the-application-instance-please':
```
andrewgodwin: I'm not too keen on this - if it's testing code, let's try to keep it in the testing code rather than the main adapter!

PR #113, opened 2019-08-11
Allow an application to instruct the server to shutdown

pgjones:
This adds to the lifespan specification that the lifespan.shutdown
message may also be sent from the application to the server to trigger
a graceful shutdown.

This is something I've been asked about, in https://github.com/pgjones/quart/issues/76, and something that seems to be supported in some WSGI servers, [Werkzeug](https://werkzeug.palletsprojects.com/en/0.14.x/serving/#shutting-down-the-server).

### DIFF
--- a/specs/lifespan.rst
+++ b/specs/lifespan.rst
@@ -92,6 +92,9 @@ Shutdown
 Sent when the server has stopped accepting connections and closed all
 active connections.
 
+Alternatively this is sent by the application when it wants the server
+to gracefully shutdown.
+
 Keys:
 
 * ``type`` (*Unicode string*) --  ``"lifespan.shutdown"``.
@@ -123,6 +126,8 @@ Keys:
 Version History
 ===============
 
+* 3.0 (2019-08-11): Allow applications to send ``lifespan.shutdown``
+  messages to a server to initiate shutdown of the server.
 * 2.0 (2019-03-04): Added startup.failed and shutdown.failed,
   clarified exception handling during startup phase.
 * 1.0 (2018-09-06): Updated ASGI spec with a lifespan protocol.


davidism:
What is a valid use case for wanting to do this? Werkzeug's thing is completely custom, and causes confusion constantly since users can't do the same thing with production servers.

pgjones:
The use case seems to always be from a desire to shut down the server via a HTTP request i.e. initiate a remote shutdown. 

I think this wouldn't create confusion as if it were to be accepted ASGI servers that advertise a lifespan spec_version >= 3.0 would support it (the app can check when the shutdown is called).

andrewgodwin:
Hmm, this feels like it's a little bit of a stretch - putting my personal view aside that graceful shutdown is an antipattern, wouldn't we be better served catching `SystemExit` or similar properly?

pgjones:
I think the case for this hasn't been made, so I'm happy to close.

parity3:
> putting my personal view aside that graceful shutdown is an antipattern

Granted, I have no experience creating an API spec, and my framework experience is limited. But off the top of my head I can think of, oh, 8 reasons why graceful shut down is a thing people need from anywhere, and is within scope of ASGI.

- It is a really good way to catch programming bugs. When certain state machines (ie web servers) have been running for a long time, exercising a shutdown can expose bugs with parts of the process which may have become silently unresponsive at some point in the lifecycle. One could argue that's what health checks are for, but this is a poor man's healthcheck. This is way easier and catches more than any healthcheck could.
- Test cases need this anyway if they want to be able to use a higher level API to manage the server
- Processes (and ASGI server implementations) can leave temp files that would cause extra burden on sys admins if left in an unclean state. Shutting down via an API is a better way to handle long term graceful shutdown for these files as well as other services. Yes, it is more work because one would need to clean these anyway if an ungraceful shutdown happened, but that cleanup would have had to be done a lot more asynchronously and would require extra separate code/logic.
- for any users unaware of the server being shut down, an API can expose that the shutdown was initiated purposefully by some user/process, vs a result of a bug or dropped network connection / etc
- Just because it is hard and most people / library authors still shoot themselves in the foot doesn't mean we should ignore shut down entirely
- All implementations already have the ability to shut down the server in some form anyway. One could argue that exposing it to the API is just a formality, but it can serve the ecosystem by providing first class support and uniformity for it. It can also encourage people to think about the true lifecycle of their service from top to bottom.
- if asgi is still in "competition" with another API, or with the possibility of no API at all, then providing what a portion of users expect from the API (ie the ability to shut down without sys.exit / TERM / KILL) would serve the community even when, in some cases, one could argue that supporting it only serves to create even more cornercases programmers would have to address when dealing with different implementations or framework-version-upgrades.
- not having a shutdown API may have quality of service implications for things like outside ssh connections, connection pools, etc that could be more explicit about resource unlocking or provide more flexibility with different failure modes, resource availability latency, etc

Can I get some rationale here besides the obvious benefits of skipping the shutdown-from-request-path option?
- It's really hard / near impossible to time a graceful shutdown with multiple persistent connections to the server. More state is required in the app / handlers.
- adding to the API when it's possible to get by without it can cause more burden on implementers and slow the adoption of async in general
- Since systems need to handle all cases including ungraceful shutdown, gracefully, then having graceful shutdown at all is redundant
- graceful shutdown may prevent people from testing the ungraceful shutdown path in tests / real world, ironically defeating its own goals
- load balancers / healthcheckers work to redirect traffic during deploys / failure modes at the reverse proxy level so it's completely out-of-path for a production system. Especially in a container-driven world, just trash the container and spin another up; nothing's truly broken unless that can't happen
- graceful shutdown can still happen; it doesn't need a formal API; just catch the INT/TERM/QUIT signal and stay running until services have shut down

Which camp are we in, mostly? I was just surprised that a spec would have the concept of a lifecycle but not expose a shutdown to 'itself'.

andrewgodwin:
The reason this was closed was not "graceful shutdown is bad", it's that Python already has mechanisms for all this (`SystemExit`, `atexit`, etc.). I don't see a need to add it into a webserver spec.

pgjones:
Should the specification say something like, `BaseException`s (note that asyncio's CancelledError derives from Exception in 3.8) raised by the application should result in the server gracefully shutting down and must result in server exiting? I think servers are free to catch and ignore all application errors at the moment.

parity3:
Ok, this will remain the responsibility of the framework implementer / sysadmin. This is a bit off-topic, but for googlers, this is how I'm achieving shutdown-from-request with a hypercorn - quart-trio setup:
```python
# raising Shutdown will just be caught and logged, and not shutdown
# from hypercorn.utils import Shutdown
from quart.wrappers.response import DataBody
from hypercorn import Config
app = QuartTrio(__name__)

class AfterDataBody(DataBody):
    def __init__(self, data: bytes, do_stuff_callback: callable):
        super().__init__(data)
        self.do_stuff_callback = do_stuff_callback

    async def __aexit__(self, exc_type: type, exc_value: BaseException, tb: TracebackType) -> None:
        self.do_stuff_callback()

@app.route('/shutdown')
async def shutdown():
    return AfterDataBody(json.dumps({'status': 'OK'}).encode(), current_app.shutdown_event.set)

app.shutdown_event = trio.Event()
from hypercorn.trio.run import worker_serve
config = Config()
await worker_serve(app, config, shutdown_trigger=app.shutdown_event.wait)
```

PR #123, opened 2019-09-10
Clarify Pseudo header treatment

pgjones:
Specifically that Pseudo headers are not an application concern, but
rather are a concern of the server (as they are specific to the
protocol).

Additionally this requires an ASGI server to create a host
header. Doing so ensures that ASGI applications can continue to assume
a host header exists.

Note there can only be one host header which should be at the start as
per RFC-7230, https://tools.ietf.org/html/rfc7230#section-5.4.


pgjones:
Ace, thanks again :). Now to implement this...

PR #125, opened 2019-09-11
tests: reduce/remove sleep

blueyed:


andrewgodwin:
Those last three sleeps you removed should stay there - they're meant to ensure the tests break and don't just run in-order by chance.

blueyed:
> they're meant to ensure the tests break and don't just run in-order by chance.

I'm not getting this.. tests are run in order anyway, and it is only testing "self" etc, nothing that needs to sleep before returning something (which indicates success).
Am I missing something?

andrewgodwin:
Well, they're synchronous sleeps inside the `sync_to_async` wrapper, designed to trigger the `PYTHONASYNCIODEBUG` alerts if they somehow leak out and get run in an event loop rather than a thread.

I believe that triggers on anything that takes over 100ms, so reducing them from 1 to 0.2 seconds would still achieve the same result.

andrewgodwin:
Closing due to inactivity.

PR #126, opened 2019-09-15
Include regression tests in MANIFEST.in to include them in PyPI tarball

kmosiejczuk:
In OpenBSD we like to gave the regression tests for testing and making sure we haven't broken things when updating other ports. 

This will include those tests in the PyPI tarball for future releases.


andrewgodwin:
Sounds good 👍

PR #137, opened 2019-11-19
Add support for naive introspection of wrapped sync_to_async functions

### LINKED ISSUE #136: Support naive introspection for wrapped functions
It does not seem like the @sync_to_async decorator can currently be used as part of a nested decorator chain.

For example, trying to use the decorator as part of a pytest fixture:

```python
@pytest.fixture
@sync_to_async
def my_test_fixture():
    pass
```

...fails with `AttributeError: 'SyncToAsync' object has no attribute '__name__'`

Aside from testing, this probably limits the range of scenarios that the decorator can be used for in general.

jcass77:
Fixes #136.

### DIFF
--- a/asgiref/sync.py
+++ b/asgiref/sync.py
@@ -198,6 +198,7 @@ class SyncToAsync:
 
     def __init__(self, func, thread_sensitive=False):
         self.func = func
+        functools.update_wrapper(self, func)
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Thanks!

PR #148, opened 2020-03-05
[RFC] Propagate changes to contextvars "upstream" for SyncToAsync

basepi:
This PR changes `SyncToAsync` to check if downstream code made changes to any contextvars, and propagates those changes back to the current context, so they'll be available for future downstream consumers.

### Background

I work on Elastic's APM team, and our [python agent](https://github.com/elastic/apm-agent-python) uses `contextvars` to keep track of the running transaction object during a request in Django. We create that transaction object [in our `request_started` signal handler](https://github.com/elastic/apm-agent-python/blob/11c64362fe5189ee30936eb8791b8e1368371dd0/elasticapm/contrib/django/apps.py#L147). The reason we do it there instead of in our middleware is so we can measure time spent in the resolver and any middleware that execute before ours.

However, for ASGI, that transaction is no longer present in `contextvars` because it is set in the [copy](https://github.com/django/asgiref/blob/911215ca81a6956df4de7e282f954654443b1fcd/asgiref/sync.py#L241) that `SyncToAsync` provides to the executor. ([PR where this was added](https://github.com/django/asgiref/pull/73))

This PR checks for any changes to the context made by the downstream code, and calls `.set()` for each of these `ContextVar` instances so those changes are "saved" for future downstream calls.

***

I ran the tests and everything passed. I can't think of any adverse side effects from this change, but I'm fairly new to ASGI in Django, thus the `[RFC]`. Thanks in advance for your comments!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ __pycache__/
 .cache
 .eggs
 .python-version
+.pytest_cache/
+.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ __pycache__/
 *~
 .cache
 .eggs
+.python-version
--- a/asgiref/sync.py
+++ b/asgiref/sync.py
@@ -265,7 +265,10 @@ class SyncToAsync:
             # Check for changes in contextvars, and set them to the current
             # context for downstream consumers
             for cvar in context:
-                if cvar.get() != context.get(cvar):
+                try:
+                    if cvar.get() != context.get(cvar):
+                        cvar.set(context.get(cvar))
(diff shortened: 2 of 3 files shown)


andrewgodwin:
Lovely, this seems like a great addition. I know we don't have any contextvars tests at all right now, but do you think you could add a quick one to verify this so we don't stop it working by mistake in future?

basepi:
Definitely. I'll do this today.

basepi:
Done!

andrewgodwin:
Awesome, thanks a lot! I'll get it out in a release soonish.

PR #152, opened 2020-03-19
New implementation of local.Local

### LINKED ISSUE #144: Using asgiref.local.Local with gevent leads to memory leaks
We use Django with Gunicorn and gevent workers. Since updating to Django 3, we started seeing some reasonably severe memory leaks in our web processes.

I bisected one of the leaks I could reproduce locally to https://github.com/django/django/commit/a415ce70bef6d91036b00dd2c8544aed7aeeaaed.

In particular, it seems to be caused by the switch from `threading.local` to  `asgiref.local.Local`.

I investigated a fair bit. There seem to be a few things going on:

- [this `isinstance(key, threading.Thread)` test](https://github.com/django/asgiref/blob/master/asgiref/local.py#L77) evaluating to `False` 

  (Note: `else:` is not handled there at all.)
- [`key.is_alive()`](https://github.com/django/asgiref/blob/master/asgiref/local.py#L78) also evaluating to `False` (this is after working around the previous problem)

(Also, I'm not sure how gevent-friendly `CLEANUP_INTERVAL = 60` is.)

It could be argued that gevent-related bugs are a problem with gevent. However, I think this problem is quite severe and also not easily noticed as it's mostly a silent problem.

While any fixes (here or in gevent) would be appreciated, it would be good to document the incompatibility too.

(I tested gevent 1.4.0 and 1.5a3.)

Thanks

andrewgodwin:
This one stores on threads/tasks directly, much like the implementation in the standard library. This removes the need for garbage-collection of local storage, and should be more compatible with libraries that override `threading`, like `gevent`.

Fixes #144.

### DIFF
--- a/asgiref/local.py
+++ b/asgiref/local.py
@@ -1,7 +1,5 @@
-import asyncio
 import sys
 import threading
-import time
 
 
 class Local:
@@ -32,10 +30,8 @@ class Local:
     CLEANUP_INTERVAL = 60  # seconds
 
     def __init__(self, thread_critical=False):
-        self._storage = {}
-        self._last_cleanup = time.time()
-        self._clean_lock = threading.Lock()
         self._thread_critical = thread_critical
+        self._attr_name = "_asgiref_local_impl_%s" % id(self)
 
     def _get_context_id(self):
         """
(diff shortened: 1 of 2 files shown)


adamchainz:
I've tested this with https://github.com/scoutapp/scout_apm_python/ , where we use it for state tracking - all seems good.

andrewgodwin:
Great. I'll likely roll this out in a release once we get confirmation it also fixes the django connection threading issue.

PR #158, opened 2020-05-04
Fix memory leak in Local

kmichel-sereema:
The _context_refs array of weak references was only freed
when the whole Local object was deleted: even when the
referenced objects weren't there anymore, the array of
empty refs was still there.

This is a problem in django where Local() objects are kept
at the global scope in multiple places and never cleared.

### DIFF
--- a/asgiref/local.py
+++ b/asgiref/local.py
@@ -35,7 +35,7 @@ class Local:
     def __init__(self, thread_critical=False):
         self._thread_critical = thread_critical
         self._thread_lock = threading.RLock()
-        self._context_refs = []
+        self._context_refs = weakref.WeakSet()
         # Random suffixes stop accidental reuse between different Locals,
         # though we try to force deletion as well.
         self._attr_name = "_asgiref_local_impl_%s_%s" % (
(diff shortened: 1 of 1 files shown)


adamchainz:
Nice work. Is there any kind of regression test we could add?

kmichel-sereema:
I had a production server where memory usage was visibly growing over a few days.

I found the exact issue by using [tracemalloc](https://docs.python.org/3/library/tracemalloc.html) and hitting repeatedly a local instance for a few hours, local was at the top of the tracemalloc stats with over 20Mb used for the array.

I was thinking about adding tests but a soak test like I did takes a few hours. I tried to write a short test by comparing tracemalloc snapshots before and after using Local only once but the snapshots are not clean (the threading module leaves stuff for instance). I'm not sure comparing and ignoring differences from other modules would really test the absence of memory leaks.

andrewgodwin:
Nice - good job spotting that one. I had a go at trying to find leaks when testing this small-scale but that's pretty difficult to do reliably.

I don't mind not having tests for this; weakref is already a sign to me that says "hey, we're being super careful here about memory leakage", so I would hope nobody wanders in here without considering this.

That said, if you do want to work on a test, I would very happily merge it.

adamchainz:
Yes nice work on the investigation! Yes it's hard to write a test proving the absence of memory leaks :/

PR #161, opened 2020-05-25
Move setup information to declarative setup.cfg

jdufresne:
Use a declarative syntax to avoid mixing code and configuration. Simplifies handling of long description and version by reducing some boilerplate.

For details on this setuptools feature, see:
https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files

Django made the same move to `setup.cfg` in: https://github.com/django/django/commit/85efc14a2edac532df1a9ad4dd9b6d4a4dcf583e


setup.cfg
```diff
@@ -1,3 +1,38 @@
...
+    License :: OSI Approved :: BSD License
+    Operating System :: OS Independent
+    Programming Language :: Python
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3 :: Only
+    Programming Language :: Python :: 3.5
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Topic :: Internet :: WWW/HTTP
```
adamchainz, line 23: Whilst we're at it, we could add some project URL's?

```suggestion
    Topic :: Internet :: WWW/HTTP
project_urls =
    Documentation = https://asgi.readthedocs.io/en/latest/
    Further Documentation = https://docs.djangoproject.com/en/dev/topics/async/#async-adapter-functions
    Changelog = https://github.com/django/asgiref/blob/master/CHANGELOG.txt
```

setup.cfg
```diff
@@ -1,3 +1,38 @@
+[metadata]
+name = asgiref
+version = attr: asgiref.__version__
+url = http://github.com/django/asgiref/
```
adamchainz, line 4:
```suggestion
url = https://github.com/django/asgiref/
```

setup.py
```diff
@@ -1,44 +1,3 @@
...
-        'Programming Language :: Python',
-        'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.5',
-        'Programming Language :: Python :: 3.6',
-        'Programming Language :: Python :: 3.7',
-        'Programming Language :: Python :: 3.8',
-        'Topic :: Internet :: WWW/HTTP',
-    ],
-)
+setup()
```
davidism, line 3: If you want GitHub's dependency/security features to continue working for this project and projects that use it, you need to at least leave the `name` in `setup()`. (`install_requires` as well, but that's not applicable here.)

adamchainz: All my projects use `setup.cfg` and get security alerts - not sure if they're passing on to others though.

Have you got a source for this?

davidism: It's project that depend on this that wouldn't get alerts, because GitHub's dependency graph feature only understands `setup.py`, not `setup.cfg`. This has been my experience with all the Pallets projects.

jdufresne:
Thanks for those suggestions @adamchainz, I have applied them.

setup.py
adamchainz: I had a look around and guessed it's parsed by their semantic, created this issue: https://github.com/github/semantic/issues/562

andrewgodwin: OK, I'd like to preserve compatibility with that feature, so it's worth moving the `name` back in, I think.

adamchainz:
We should also add `name` back in on Django, for the time being.

setup.cfg
```diff
@@ -1,3 +1,42 @@
...
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3 :: Only
+    Programming Language :: Python :: 3.5
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Topic :: Internet :: WWW/HTTP
+project_urls =
+    Documentation = https://asgi.readthedocs.io/
+    Further Documentation = https://docs.djangoproject.com/en/dev/topics/async/#async-adapter-functions
```
jdufresne, line 26: I think this should use `stable` instead instead of `dev`.

jdufresne:
Thanks for the reviews, feedback, and followup all! I have added the `name` argument back to `setup.py`.

andrewgodwin:
Thanks for this!

PR #162, opened 2020-05-25
Remove executable bit from setup.py

jdufresne:
The file has no shebang, so it should not be executed directly.


adamchainz:
Well spotted

PR #164, opened 2020-06-01
📝 Minor wording update: Introduction

tiangolo:
This PR is a minor wording update for consistency.

Making the types of the parameters passed to an application a bit more explicit and making more explicit that the functions are "asynchronous callables", to use the same wording used in other parts of the spec and in the Python docs.

It also updates one of the example event messages to show what could be received in a `receive` event message. Currently, the text refers to `receive` but the example is for `send`. As there are other examples for `send` I thought it would be better to have one example for `receive`.

---

This is part of several PRs I'm making while reading and re-reading the spec several times.

The main purpose is to try and update the parts that contained text based on the 2-callable version (`2.0`) and to clarify a bit more the parameters passed to an ASGI app and types and direction of events.

### DIFF
--- a/docs/introduction.rst
+++ b/docs/introduction.rst
@@ -29,17 +29,19 @@ receiving WebSocket frames) can't trigger this.
 How does ASGI work?
 -------------------
 
-ASGI is structured as a single, asynchronous callable. It takes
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks for merging! :tada:

PR #165, opened 2020-06-01
📝 Minor wording update: ASGI spec

tiangolo:
This PR is a minor wording update of the spec.

It mainly deals with rewording the parts that used to refer to the 2-callable design in ASGI 2.0.

There's a small text update to show that although `asyncio` is the canonical implementation, it is not required by the spec, what is required is `async` / `await` coroutine support.

It clarifies a bit the types and usage of the arguments passed to an ASGI app.

There's also a small clarification to make explicit the *connection scope* as a technical term used by the spec to include the lifetime of a connection and the `scope` parameter passed to the app with *connection scope* information.

There are also some other minor wording updates mostly for consistency with the rest of the spec and to try and reduce any possible ambiguity.

---

This is part of several PRs I'm making while reading and re-reading the spec several times.

The main purpose is to try and update the parts that contained text based on the 2-callable version (2.0) and to clarify a bit more the parameters passed to an ASGI app and types and direction of events.

### DIFF
--- a/specs/asgi.rst
+++ b/specs/asgi.rst
@@ -53,13 +53,16 @@ ASGI consists of two different components:
 - A *protocol server*, which terminates sockets and translates them into
   connections and per-connection event messages.
 
-- An *application*, which lives inside a *protocol server*, is instantiated
-  once per connection, and handles event messages as they happen.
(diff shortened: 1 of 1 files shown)


tiangolo:
Thanks for merging! :tada:

Issue #110, opened 2019-08-03
ASGI resources list


florimondmanca:
Hey!

I'm opening this discussion after originally poking the idea over the [Encode discussion board](https://discuss.encode.io/t/discoverability-of-the-asgi-ecosystem/729/2).

## Context

ASGI has been establishing itself as a common interface enabling interoperability across a whole range of things in the Python async web space, including:

- Apps ([Datasette](https://github.com/simonw/datasette))
- Frameworks (Starlette, and soon Django!)
- Middleware ([sentry-asgi](https://github.com/encode/sentry-asgi), [timing-asgi](https://github.com/steinnes/timing-asgi))
- Wrappers ([mangum](https://github.com/erm/mangum))
- Libraries (this very project)

There are also all sorts of non-code resources out there, like talks and articles.

But, as the ASGI ecosystem is growing, I personally feel like it's hard to keep up on which things are available.

## Listing ASGI resources

In this context, perhaps it could be worth setting up some sort of collaborative list of ASGI resources. This would allow better discoverability and more wide-spread usage of ASGI-related projects, and help getting a grasp of the *scale* of the ecosystem.

It seems a collaboratively-grown static list (e.g. an ASGI [awesome list](https://github.com/sindresorhus/awesome) or an [Are we … yet?](https://wiki.mozilla.org/Areweyet) list) would fill this needs quite well already.

Just to be clear: the ASGI list wouldn't list *any* async resources (the [awesome-asyncio](https://github.com/timofurrer/awesome-asyncio) is already very well suited for that), but only those that are directly related to ASGI, such as those I listed earlier.

## Location

Technically the list *could* be in the ASGI documentation itself, but I feel like the ASGI docs/spec should be part of the list, not the other way around. Giving the list its own space such as a GitHub repo is probably a more sustainable approach.

Digging further into this, maybe now would be a good time to *give ASGI its own space*, e.g. a `python-asgi` GitHub org. I suppose it'd contain this repository and the ASGI list at first (and maybe turn into some sort of umbrella organization for projects?).

Anyway, happy to read everyone's opinion on this. I'd love to help setup the list if this sounds like a good idea at all.

cc @tomchristie, @andrewgodwin, @simonw (who else?)

andrewgodwin:
So, we started a list of ASGI things over at https://asgi.readthedocs.io/en/latest/implementations.html, in the vein of this, but I agree it could maybe be a bit more open/easy-to-edit.

[...]

PR #168, opened 2020-06-01
📝 Add FastAPI to Implementations

tiangolo:
This adds FastAPI to the ASGI implementations :tada: :rocket:

### DIFF
--- a/docs/implementations.rst
+++ b/docs/implementations.rst
@@ -48,6 +48,16 @@ original driving force behind the ASGI project. Supports HTTP and WebSockets
 with Django integration, and any protocol with ASGI-native code.
 
 
+FastAPI
+-------
+
+*Beta* / https://github.com/tiangolo/fastapi
+
+FastAPI is an ASGI web framework (made with Starlette) for building web APIs based on
+standard Python type annotations and standards like OpenAPI, JSON Schema, and OAuth2.
+Supports HTTP and WebSockets.
+
+
 Quart
 -----
 
@@ -65,13 +75,3 @@ Starlette
 
 Starlette is a minimalist ASGI library for writing against basic but powerful
 ``Request`` and ``Response`` classes. Supports HTTP.
-
-
-FastAPI
--------
-
-*Beta* / https://github.com/tiangolo/fastapi
-
-FastAPI is an ASGI web framework (made with Starlette) for building web APIs based on
-standard Python type annotations and standards like OpenAPI, JSON Schema, and OAuth2.
-Supports HTTP and WebSockets.
--- a/docs/implementations.rst
+++ b/docs/implementations.rst
@@ -73,5 +73,5 @@ FastAPI
 *Beta* / https://github.com/tiangolo/fastapi
 
 FastAPI is an ASGI web framework (made with Starlette) for building web APIs based on
-standard Python type annotations, based on standards like OpenAPI, JSON Schema, and
-OAuth2. Supports HTTP and WebSockets.
+standard Python type annotations and standards like OpenAPI, JSON Schema, and OAuth2.
+Supports HTTP and WebSockets.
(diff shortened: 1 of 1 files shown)


docs/implementations.rst
```diff
@@ -65,3 +65,13 @@ Starlette
 
 Starlette is a minimalist ASGI library for writing against basic but powerful
 ``Request`` and ``Response`` classes. Supports HTTP.
+
+
+FastAPI
```
adamchainz, line 70: The list is in alphabetical order and I presume we'd want to keep it that way, so move above Quart

tiangolo:
Not sure if you want to mention Starlette-based frameworks, that's your call :shrug: ...I can add them too if you want. I think Sanic also supports ASGI now, that could probably go in there too.

Would you like me to add them or would you prefer to keep only Starlette here?

erm:
There was some discussion [here](https://github.com/django/asgiref/issues/110) touching on this topic more broadly. I think this PR is revealing how difficult it may be to sustain the implementation list in its current form, specifically the lack of a definitive criteria for inclusion in this case.

tiangolo:
Thanks @erm !

It makes sense, in that case, I guess it's probably fine to just close this issue.

And maybe we could add a pointer to https://github.com/florimondmanca/awesome-asgi from here as it has many other related resources.

tomchristie:
Personally I think it'd make sense to include FastAPI, Responder, Flama yeah - they're all significantly different from each other, and significantly different from working with Starlette as a framework, so it'd be a clear thumbs up from me. I also don't think they necessarily need to mention on this resource that they happen to use Starlette. (Tho it's always appreciated on their own docs when they do)

adamchainz:
I think it might be more sustainable to remove the list here and point to awesome-asgi only, otherwise there will be two community lists of implementations.

tiangolo:
I guess we can close this PR as I imagine those changes would be in a different one, right?

On the other hand, I'm not sure how would you like to structure it after removing the "implementations" section and where to put the link to awesome-asgi. I guess it might be simpler if you add it directly in a way you're comfortable with... otherwise, would you like me to propose something?

andrewgodwin:
Let's at least add FastAPI in for now (could you fix the ordering so it matches the existing alphabetical ordering?)

I agree it would be nice to just have awesome-asgi, but we can do this quickly now, and not block small easy changes on a harder, bigger change.

tiangolo:
Cool!

Sure, done :heavy_check_mark:

andrewgodwin:
Thanks! We should definitely look at awesome-asgi too, though.

Kitty2014:
@tiangolo  Whether FastAPI can support GRPC implementation?

tiangolo:
@Kitty2014 it's probably better to ask it in an issue in the FastAPI repo to not pollute much here.

But the short answer is, not really.

FastAPI is designed for web servers/APIs, it is based on Starlette, Starlette is a web framework/toolkit, so it uses the ASGI sub-specs for HTTP and WebSockets. All that is handled and run by an ASGI server like Uvicorn that listens on the web ports and runs the application.


---

Now, about gRPC, ASGI:

GRPC would probably first need an ASGI sub-spec, then an implementation server. And then probably a framework for GRPC.

But gRPC already has all its own ideas, framework, and tools. If I remember correctly, the spec, communication, and everything are heavily based on protocol buffers, which in turn, are heavily based on a C++ implementation, that includes some quirks, for example, many object attributes can be both scalar values and sequences at the same time, and it heavily depends on classes (i.e. no `dict`s or `list`s) but without the benefit of classes, like static type checks, completion, etc, all that ends up being quite awkward in Python (I don't know if all that makes more sense in C++).

Also, if I remember correctly it's all quite tightly-coupled, and pretty much expects you to use their provided packages, so, being honest, if you have a simple problem and want to do gRPC you might be better off using those than through the whole route of the ASGI sub-spec. Of course, it would be totally feasible, but probably a lot more work.

---

Now, if what you want is a way to create some type of remote server that can easily communicate in a binary format with clients, then you can probably use FastAPI (or any other ASGI framework) with an ASGI middleware like https://github.com/florimondmanca/msgpack-asgi

PR #171, opened 2020-06-05
Propagate and restore contextvars in async_to_sync

### LINKED ISSUE #170: Excessive use of asgiref.local and sync_to_async breaks other context local implementations
I am writing on behalf of my employer, [sentry.io](https://sentry.io).

I am not sure whether this issue should be filed against Django or asgiref. Starting with Django 3.1, signal handlers for `got_request_exception` are executed with the wrong `contextvars` context when running as part of an ASGI request. This can be easily demonstrated by wrapping the Django ASGI app in a middleware that stores something on the context, and trying to retreive that same value from that signal handler.

This is how I became aware of Django's custom context local implementation that is shipped as part of asgiref. I understand why you do the things you do, but consider that this will break a lot of tracing tooling for Python that may only be orthogonal to Django or ASGI. [The situation around context locals is already awful enough](https://docs.sentry.io/platforms/python/contextvars/), it would be terrible if Sentry and tools like it had to implement a codepath specifically for Django 3 applications, or anything that uses asgiref.local.

I can see the following options:

1. avoid releasing Django 3.1 with whatever changes just happened that made errorhandlers execute in the wrong `contextvars` context. Considering that Django 3.1 is already in alpha I would at least consider this as a stopgap measure.
2. modify asgiref.sync_to_async and its inverse to "drive" contextvars if they are available such that contextvars always behave identically to asgiref.local
3. drop asgiref.local entirely and swap it out for the `aiocontextvars` package on PyPI, as the primary purpose of implementing asgiref.local from scratch seemed to have been to support Python 3.6. One would still have to implement 2) for this to work, but the advantage here is that aiocontextvars seems to be significantly more popular across frameworks than asgiref.local.

Those options are not mutually exclusive.

Thanks for considering.

untitaker:
sync_to_async already makes sure that contextvars are properly
propagated, but async_to_sync does not attempt to do this at all.

In Django 3.1, this fixes the behavior of got_request_exception which
would otherwise be called in a different context than before_request or
after_request under the following circumstances:

1. An ASGI request has been made
2. There is at least one sync middleware

Fix #170

### DIFF
--- a/asgiref/sync.py
+++ b/asgiref/sync.py
@@ -15,6 +15,29 @@ except ImportError:
     contextvars = None
 
 
+def _restore_context(context):
+    # Check for changes in contextvars, and set them to the current
+    # context for downstream consumers
+    for cvar in context:
+        try:
+            if cvar.get() != context.get(cvar):
+                cvar.set(context.get(cvar))
+        except LookupError:
+            cvar.set(context.get(cvar))
+
+
+def _awaitable_with_context(awaitable, context):
+    gen = awaitable.__await__()
+
+    while True:
+        try:
+            chunk = context.run(next, gen)
+        except StopIteration:
+            break
+
+        yield chunk
+
+
 class AsyncToSync:
     """
     Utility class which turns an awaitable that only works on the thread with
@@ -66,6 +89,10 @@ class AsyncToSync:
                     "You cannot use AsyncToSync in the same thread as an async event loop - "
                     "just await the async function directly."
                 )
+
+        if contextvars is not None:
+            context = contextvars.copy_context()
+
         # Make a future for the return information
         call_result = Future()
         # Get the source thread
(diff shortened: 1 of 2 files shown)


asgiref/sync.py
```diff
@@ -15,6 +15,29 @@
     contextvars = None
 
 
+def _restore_context(context):
```
andrewgodwin, line 18: Since these two functions are only used by AsyncToSync, could you move them underneath it as classmethods or similar?

untitaker: `_restore_context` is shared across the two classes, what's your preference there?

untitaker:
@andrewgodwin could you take a look at this again?

andrewgodwin:
I will when I get the time and mental energy to fully review it again. Hopefully this week.

untitaker:
Sounds good, thanks and sorry for being pushy.

andrewgodwin:
Thanks - you're right to leave the restore_context function outside, missed it was being used by both sides. I've merged it in, and I'll issue a release shortly.

PR #172, opened 2020-06-07
Added fastAPI to the list of app frameworks

rednafi:
[FastAPI](https://github.com/tiangolo/fastapi) is a modern, high-performance framework to develop APIs. It's based on [Starlette](https://github.com/encode/starlette) and uses [Pydantic](https://github.com/samuelcolvin/pydantic/) for automatic data validation from type hints. Added that to the list of frameworks.


adamchainz:
Duplicate of #168

PR #175, opened 2020-06-16
Rewrite how context is propagated in AsyncToSync

### LINKED ISSUE #170: Excessive use of asgiref.local and sync_to_async breaks other context local implementations

untitaker:
Fix #170 again. This has been run against the Django master testsuite.

Not entirely sure if this performs. Do we have a benchmark suite for this (in Django)?

### DIFF
--- a/asgiref/sync.py
+++ b/asgiref/sync.py
@@ -79,7 +79,11 @@ class AsyncToSync:
                 )
 
         if contextvars is not None:
-            context = contextvars.copy_context()
+            # Wrapping context in list so it can be reassigned from within
+            # `main_wrap`.
+            context = [contextvars.copy_context()]
+        else:
+            context = None
 
         # Make a future for the return information
         call_result = Future()
@@ -99,11 +103,8 @@ class AsyncToSync:
         # in this thread.
         try:
             awaitable = self.main_wrap(
-                args, kwargs, call_result, source_thread, sys.exc_info()
+                args, kwargs, call_result, source_thread, sys.exc_info(), context
             )
-
-            if contextvars is not None:
-                awaitable = self._awaitable_with_context(awaitable, context)
 
             if not (self.main_event_loop and self.main_event_loop.is_running()):
                 # Make our own event loop - in a new thread - and run inside that.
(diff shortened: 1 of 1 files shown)


andrewgodwin:
We don't have a benchmark suite, sadly. I'll pull this branch in locally and double-check it against the Django test suite before I merge this time just to be sure :)

andrewgodwin:
Great, thanks for getting this turned around so fast!

apollo13:
Is there a performance reason to use a list to wrap the context as opposed to a class/dataclass?

untitaker:
@apollo13 no particular reason. I think the main performance impact comes from the _restore_context invocations, each of which are `O(n)` with `n` being the amount of contextvars in the process

PR #176, opened 2020-06-18
Fix a crash in Local.__del__ due to buggy WeakSet impl

untitaker:


adamchainz:
Is a test possible? Is there a Python bug to link to in a comment as it might get fixed in a future version?

untitaker:
I couldn't find a python bug report. I can file one later.

Our options for tests are constrained to mocking weakset to always raise a TypeError, not sure how to simulate interpreter shutdown in the tests

adamchainz:
The mocking approach sounds reasonable to me.

Thanks for all your work here!

andrewgodwin:
@untitaker Is this related to the continuing saga of Django tests breaking mentioned in #170? I see it relates to the only other change in 3.2.8 so I'm guessing so.

untitaker:
It relates to a change in 3.2.8 unrelated to anything context propagation:  edd0570

andrewgodwin:
Yup, that's the one I meant. If you get a mocked test in with a fake TypeError raise we can land and release this pronto.

untitaker:
Here you go. Not entirely sure if that test is any good but oh well.

andrewgodwin:
Just for clarification - is this bug in WeakSet only in certain Python versions?

untitaker:
I think it's been there since 2.7, the code looks almost exactly the same: https://github.com/python/cpython/blob/507bd8cde60ced74d13a1ffa883bb9b0e73c38be/Lib/_weakrefset.py#L59

I haven't tested though

andrewgodwin:
Alright, this is out in `3.2.10`. Hopefully that's the last of the damage from 3.2.8 :)

PR #183, opened 2020-07-14
fix encoding error

### LINKED ISSUE #182: WSGI Error in non-ascii path
https://github.com/encode/starlette/issues/997

I noticed that the ASGI document clearly states that utf8 encoding is required, but WSGI seems to require latin-1 encoding.

https://www.python.org/dev/peps/pep-3333/#a-note-on-string-types
https://asgi.readthedocs.io/en/latest/specs/www.html#connection-scope

abersheeran:
fix #182 

The latin1 encoding is required in WSGI, and the utf8 encoding in ASGI.

### DIFF
--- a/asgiref/wsgi.py
+++ b/asgiref/wsgi.py
@@ -55,8 +55,8 @@ class WsgiToAsgiInstance:
         """
         environ = {
             "REQUEST_METHOD": scope["method"],
-            "SCRIPT_NAME": scope.get("root_path", ""),
-            "PATH_INFO": scope["path"],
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Looks good for the fix - any chance you could get a test in for this? It's just the sort of thing that we might break by mistake otherwise.

abersheeran:
I will write a test for it. Wait a moment.

andrewgodwin:
Thanks so much for the speedy pull request!

PR #188, opened 2020-08-14
Title level inconsistent

euri10:
The rst seems a little bit off for the last 2 titles in `www.rst`


andrewgodwin:
Thanks for catching that!

PR #189, opened 2020-08-16
Anchor links in docs are sometimes not "nice"

euri10:
this happens twice in the docs (I didnt see more) and it happens when titles are the same, in that case sphinx generates an anchor link for the 2nd one that looks like this:

https://asgi.readthedocs.io/en/latest/specs/www.html#id1
https://asgi.readthedocs.io/en/latest/specs/www.html#id2

the 1st one is for the Connection scope, it's used twice for http and socket protocol, I dealt with it adding HTTP / Websocket
the 2nd one is for the receive disconnect message,  I dealt with it adding an explicit anchor reference on the 2nd one

### DIFF
--- a/specs/www.rst
+++ b/specs/www.rst
@@ -53,8 +53,8 @@ The HTTP protocol should be signified to ASGI applications with a ``type``
 value of ``http``.
 
 
-Connection Scope
-''''''''''''''''
+HTTP Connection Scope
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Makes sense - while they may not be used a whole ton, it's nice to have them better.

PR #190, opened 2020-08-17
Clarify Optional meaning

pgjones:
This clarifies that optional implies the key is missing, not that the
key is present with a value of say None (as optional means in the
Python typing context).


andrewgodwin:
:+1:

PR #191, opened 2020-09-01
Add ZeroCopy extension

### LINKED ISSUE #184: Proposal: "http.response.sendfile" extension.
Add the sendfile extension to allow ASGI programs to return response through `sendfile`.

In `scope`:

```python
"scope": {
    ...
    "extensions": {
        "http.response.sendfile": {},
    },
}
```

Response:

```python
await send({"type": "http.response.start", ...})
await send({"type": "http.response.sendfile", "file": file-obj})
```

Some related links:

https://docs.python.org/3.7/library/asyncio-eventloop.html#asyncio.loop.sendfile
https://github.com/encode/uvicorn/issues/35

abersheeran:
Close #184

### DIFF
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -78,10 +78,10 @@ and treat it as if the client had made a request.
 The ASGI server should set the pseudo ``:authority`` header value to
 be the same value as the request that triggered the push promise.
 
-Sendfile
+ZeroCopy
 ------------------
 
-Sendfile allows you to send the contents of a file descriptor to the
+ZeroCopy allows you to send the contents of a file descriptor to the
 HTTP client with zero copy. ASGI servers that implement this extension will
 provide ``http.response.sendfile`` in the extensions part of the scope::
 
@@ -92,7 +92,7 @@ provide ``http.response.sendfile`` in the extensions part of the scope::
         },
     }
 
-An ASGI framework can call ``sendfile`` by sending a message with
+The ASGI framework can call ``sendfile`` by sending a message with
 the following keys. This message can be sent at any time after the
 *Response Start* message but before the final *Response Body* message.
 
@@ -110,4 +110,9 @@ Keys:
 * ``count`` (*int*): Optional. ``count`` is the number of bytes to
   copy between the file descriptors.
 
+After calling this extension to respond, the ASGI framework itself should
+actively close the used file descriptor.
 
+In the absence of subsequent content, the ASGI framework should call
+``await send({"type":"http.response.body", "body": b"", "more_body": False})``
+to mark the end of the response. Otherwise, just continue to call ``send``.
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -78,10 +78,10 @@ and treat it as if the client had made a request.
 The ASGI server should set the pseudo ``:authority`` header value to
 be the same value as the request that triggered the push promise.
 
-ZeroCopy
+Zero Copy Send
 ------------------
 
-ZeroCopy allows you to send the contents of a file descriptor to the
+Zero copy send allows you to send the contents of a file descriptor to the
 HTTP client with zero copy. ASGI servers that implement this extension will
 provide ``http.response.sendfile`` in the extensions part of the scope::
 
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -103,7 +103,7 @@ zero copy, it should behave the same as ordinary ``http.response.body``.
 
 Keys:
 
-* ``type`` (*Unicode string*): ``"http.response.sendfile"``
+* ``type`` (*Unicode string*): ``"http.response.zerocopysend"``
 
 * ``file`` (*file descriptor object*): An opened file descriptor object
   with an underlying OS file descriptor that can be used to call ``os.sendfile``.
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -77,3 +77,37 @@ and treat it as if the client had made a request.
 
 The ASGI server should set the pseudo ``:authority`` header value to
 be the same value as the request that triggered the push promise.
+
+Sendfile
+------------------
+
+Sendfile allows you to send the contents of a file descriptor to the
+HTTP client with zero copy. ASGI servers that implement this extension will
+provide ``http.response.sendfile`` in the extensions part of the scope::
+
+    "scope": {
+        ...
+        "extensions": {
+            "http.response.sendfile": {},
+        },
+    }
+
+An ASGI framework can call ``sendfile`` by sending a message with
+the following keys. This message can be sent at any time after the
+*Response Start* message but before the final *Response Body* message.
+
+Keys:
+
+* ``type`` (*Unicode string*): ``"http.response.sendfile"``
+
+* ``file`` (*file descriptor object*): An open file descriptor object:
+  like file or socket.
+
+* ``offset`` (*int*): Optional. If this value exists, it will specify
+  the offset at which sendfile starts to read data from ``file``.
+  Otherwise, it will be read from the current position of ``file``.
+
+* ``count`` (*int*): Optional. ``count`` is the number of bytes to
+  copy between the file descriptors.
+
+
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -94,14 +94,18 @@ provide ``http.response.sendfile`` in the extensions part of the scope::
 
 The ASGI framework can call ``sendfile`` by sending a message with
 the following keys. This message can be sent at any time after the
-*Response Start* message but before the final *Response Body* message.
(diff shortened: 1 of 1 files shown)


docs/extensions.rst
```diff
@@ -77,3 +77,37 @@ and treat it as if the client had made a request.
 
 The ASGI server should set the pseudo ``:authority`` header value to
 be the same value as the request that triggered the push promise.
+
+Sendfile
+------------------
+
+Sendfile allows you to send the contents of a file descriptor to the
+HTTP client with zero copy. ASGI servers that implement this extension will
+provide ``http.response.sendfile`` in the extensions part of the scope::
```
andrewgodwin, line 86: I know it's a draft, but I'd like to make this clearer, especially as regards how it is _not_ the same as `X-Sendfile`.

docs/extensions.rst
```diff
@@ -77,3 +77,37 @@ and treat it as if the client had made a request.
...
+    "scope": {
+        ...
+        "extensions": {
+            "http.response.sendfile": {},
+        },
+    }
+
+An ASGI framework can call ``sendfile`` by sending a message with
+the following keys. This message can be sent at any time after the
+*Response Start* message but before the final *Response Body* message.
```
andrewgodwin, line 97: Should we say that it's not allowed to send a response body if you also use this?

docs/extensions.rst
```diff
@@ -77,3 +77,37 @@ and treat it as if the client had made a request.
...
+
+Keys:
+
+* ``type`` (*Unicode string*): ``"http.response.sendfile"``
+
+* ``file`` (*file descriptor object*): An open file descriptor object:
+  like file or socket.
```
andrewgodwin, line 104: I expected this to take a filename, not a file descriptor. What's the reasoning here?

davidism: I *think* they're trying to say "file-like object", maybe to allow for returning `BytesIO`? However, if this is required for regular files too, it needs to be made clear what should be closing the file and when.

abersheeran: > I expected this to take a filename, not a file descriptor. What's the reasoning here?

Using a file descriptor allows to bridge two TCP Sockets, not necessarily files.

abersheeran: > I *think* they're trying to say "file-like object", maybe to allow for returning `BytesIO`? However, if this is required for regular files too, it needs to be made clear what should be closing the file and when.

The purpose of this extension is to allow zero-copy data transfer using the sendfile feature. BytesIO should uses a normal method to transfer.

docs/extensions.rst
abersheeran: Zero-copy transmission data should not be special, it should be mixed with `http.response.body`.

andrewgodwin: Ah, interesting. That should be made very clear here, then - that servers should send this data exactly between any other `response.body` messages.

docs/extensions.rst
andrewgodwin: Yup, now I've read the appropriate asyncio source code, I see. Let's make the phrasing a bit clearer, then - maybe "a Python file object with an underlying OS file descriptor (e.g. not BytesIO)"

abersheeran: Although I don't think the "file descriptor" is not clear enough, I will change it.

docs/extensions.rst
```diff
@@ -77,3 +77,46 @@ and treat it as if the client had made a request.
 
 The ASGI server should set the pseudo ``:authority`` header value to
 be the same value as the request that triggered the push promise.
+
+Zero Copy Send
+------------------
```
andrewgodwin, line 82: Underline is wrong length

docs/extensions.rst
```diff
@@ -77,3 +77,46 @@ and treat it as if the client had made a request.
...
+
+Zero copy send allows you to send the contents of a file descriptor to the
+HTTP client with zero copy. ASGI servers that implement this extension will
```
andrewgodwin, line 85: Let's elaborate on "zero copy" - something like "zero copy (where the underlying OS directly handles the data transfer from a source file or socket without loading it into Python and out again)"

docs/extensions.rst
```diff
@@ -77,3 +77,46 @@ and treat it as if the client had made a request.
...
+provide ``http.response.sendfile`` in the extensions part of the scope::
+
+    "scope": {
+        ...
+        "extensions": {
+            "http.response.sendfile": {},
```
andrewgodwin, line 91: Can we change the extension identifier to be `http.response.zerocopysend`?

docs/extensions.rst
```diff
@@ -77,3 +77,46 @@ and treat it as if the client had made a request.
...
+        },
+    }
+
+The ASGI framework can call ``sendfile`` by sending a message with
```
andrewgodwin, line 95: Let's reword this from "call sendfile" to "initiate a zero-copy send"

docs/extensions.rst
```diff
@@ -77,3 +77,46 @@ and treat it as if the client had made a request.
...
+the following keys. This message can be sent at any time after the
+*Response Start* message but before the final *Response Body* message,
+it can be mixed with ``http.response.body``. It can also be called
```
andrewgodwin, line 98: it can -> and can

docs/extensions.rst
```diff
@@ -77,3 +77,46 @@ and treat it as if the client had made a request.
...
+* ``file`` (*file descriptor object*): An opened file descriptor object
+  with an underlying OS file descriptor that can be used to call ``os.sendfile``.
+  (e.g. not BytesIO)
+
+* ``offset`` (*int*): Optional. If this value exists, it will specify
+  the offset at which sendfile starts to read data from ``file``.
+  Otherwise, it will be read from the current position of ``file``.
+
+* ``count`` (*int*): Optional. ``count`` is the number of bytes to
+  copy between the file descriptors.
```
andrewgodwin, line 115: I think we should allow for a `more_body` key in here if it's going to be mixed with `http.response.body` - that way, there's no need to do a separate close event?

andrewgodwin (commented):
Thanks for your work on this so far - more thoughts below. I do generally like the shape of this - which is why I am being picky about wording, that's a good thing, means we're close :)

docs/extensions.rst
abersheeran: good idea.

docs/extensions.rst
abersheeran: This name continues from https://linux.die.net/man/2/sendfile and https://docs.python.org/3.7/library/asyncio-eventloop.html#asyncio.loop.sendfile

docs/extensions.rst
abersheeran: Ok.

docs/extensions.rst
abersheeran: I agree.

docs/extensions.rst
abersheeran: My English is not good. If there are any grammatical errors, please let me know. Thanks for your review.

docs/extensions.rst
abersheeran: I think it would be better to keep the name `sendfile`.

andrewgodwin: I know, but there is also an `X-Sendfile` HTTP header that is commonly used, so we cannot just call it "sendfile".

docs/extensions.rst
andrewgodwin: That's OK then, I will do grammar fixes for you once we're ready to land this!

docs/extensions.rst
abersheeran: Okay, I agree that a distinction should be made.

abersheeran:
Is there no problem here? Can I turn it into a formal PR?

docs/extensions.rst
```diff
@@ -77,3 +77,50 @@ and treat it as if the client had made a request.
...
+The ASGI framework can initiate a zero-copy send by sending a message with
+the following keys. This message can be sent at any time after the
+*Response Start* message but before the final *Response Body* message,
+and can be mixed with ``http.response.body``. It can also be called
+multiple times in one response. Except for the characteristics of
+zero copy, it should behave the same as ordinary ``http.response.body``.
+
+Keys:
+
+* ``type`` (*Unicode string*): ``"http.response.sendfile"``
```
andrewgodwin, line 106: This should match the new name above

andrewgodwin:
Of course - the way I see "draft" is that you're not ready yet, not that I am. If you want to switch it to a full PR, I think we're good to go with one small change I noted (I'll do a grammar pass for you too after it's landed).

abersheeran:
Okay, thanks for your help.

andrewgodwin:
Finally merged - thanks for your work on this and your patience!

Issue #112, opened 2019-08-08
Specify access to transport information


hashbrowncipher:
Use case: I would like to access the peer certificate sent by a TLS client within my ASGI app.

Issue: As far as I can tell, no information about the transport (including TLS session information) is available to the App.

In a fork of uvicorn, I have made a [patch](https://github.com/hashbrowncipher/uvicorn/commit/aab56c1911589375dca2cfde059d8297f5277a1d) which exposes the full `transport` object as part of the `scope` passed to the app.

```
diff --git a/uvicorn/protocols/http/h11_impl.py b/uvicorn/protocols/http/h11_impl.py
index 240cb35..cf5fd67 100644
--- a/uvicorn/protocols/http/h11_impl.py
+++ b/uvicorn/protocols/http/h11_impl.py
@@ -191,6 +191,7 @@ class H11Protocol(asyncio.Protocol):
                     "raw_path": raw_path,
                     "query_string": query_string,
                     "headers": self.headers,
+                    "transport": self.transport,
                 }
```

Within my app code, I use `scope['transport'].get_extra_info("ssl_object").getpeercert(binary_form=True)` to access the relevant information.

Feature request: Specify a way for applications to fetch information about the transport they are communicating over.

See also: https://github.com/encode/uvicorn/issues/400

andrewgodwin:
Hm, one problem with this is that the ASGI events are deliberately limited to a JSON-plus-bytes-type set of potential values; passing a full object like this is sort of difficult to specify.

Is there any prior art in WSGI apps for doing similar things? I totally see the use case for information from client certificates especially, but I'm wary of specifying it rather than leaving it to servers to experiment with and decide on what would be best.

hashbrowncipher:
It looks like uwsgi provides a nonstandard `HTTPS_CLIENT_CERTIFICATE`, with the raw bytes of the client cert. I was unable to find anything in gunicorn implementing client certificates, except for this [gist](https://gist.github.com/jmvrbanac/089540b255d6b40ca555c8e7ee484c13) where someone implemented a custom worker to access information from the client cert. So it looks like the Python ecosystem doesn't yet have much prior art with regards to accessing the TLS session information.

I'm certainly not wedded to the very wide interface suggested by my sample patch. I suggest taking a look at specifying access to some subset of [BaseTransport.get_extra_info()](https://docs.python.org/3/library/asyncio-protocol.html#asyncio.BaseTransport.get_extra_info), and moving from there.

[...]

PR #192, opened 2020-09-02
Add TLS specification

jonfoster:
Issue #112


specs/tls.rst
```diff
@@ -0,0 +1,201 @@
...
+The *connection scope* information passed in ``scope`` contains an
+``"extensions"`` key, which contains a dictionary of extensions.  Inside that
+dictionary, the key ``"tls"`` identifies the extension specified in this
+document.  The value will be a dictionary with the following entries:
+
+* ``server_cert`` (*Optional[Unicode string]*) -- The PEM-encoded conversion
+  of the x509 certificate sent by the server when establishing the TLS
+  connection.  Some web server implementations may be unable to provide this
+  (e.g. if TLS is terminated by a separate proxy or load balancer); in that
+  case this shall be ``None``.  Mandatory.
```
andrewgodwin, line 58: The way the types are written here is a little confusing if you're not familiar with the typing module - the sentence has both `Optional` and `Mandatory` in it, and the type value is not `Optional[unicode]`. May I suggest the type is instead written as `Unicode string or None` in these cases?

specs/tls.rst
```diff
@@ -0,0 +1,201 @@
...
+
+Rationale (Informative)
+-----------------------
+
+This section explains the choices that led to this specification.
+
+Providing the entire TLS certificates in ``client_cert_chain``, rather than a
+parsed subset:
+
+* makes it easier for webservers to implement, as they do not have to
```
andrewgodwin, line 125: Nit: Let's capitalise the start of each bullet point

specs/tls.rst
```diff
@@ -0,0 +1,201 @@
...
+``client_cert_chain``.  However, many ASGI applications will probably find
+that information is sufficient for their application - it provides a simple
+string that identifies the user.  It is simpler to use than parsing the x509
+certificate.  For the server, this information is readily available.
+
+There are theoretical interoperability problems with ``client_cert_name``,
+since it depends on a list of object ID names that is maintained by IANA and
+theoretically can change.  In practise, this is not a real problem, since the
+object IDs that are actually used in certificates have not changed in many
+years.  So in practise it will be fine.
```
andrewgodwin, line 195: practise -> practice

andrewgodwin (commented):
I think this looks good overall - left a few small comments, but I do generally like the design and I think it addresses most of the problems raised in the original TLS ticket.

Currently, extensions are in a separate extensions section of the docs; however, this is definitely lengthly enough to need its own one. I think I am OK leaving it in specs for now, but if you could include a cross-link from the extensions section, that'd be great.

specs/tls.rst
jonfoster: Thanks, should be fixed now.  I ran spellcheck this time, too!

specs/tls.rst
jonfoster: Done!

specs/tls.rst
jonfoster: Sure, changed.

andrewgodwin:
I'm honestly not sure if we ended up with me having to re-review it or more changes needed. I'll take a look at it sometime this month.

andrewgodwin:
Well, it took longer than a month, but on further review it looks good, so it's merged.

PR #196, opened 2020-09-16
Fixes #195: Clamp WsgiToAsgi response body using Content-Length value

### LINKED ISSUE #195: WsgiToAsgi should stop sending response body after content-length bytes
Hi,

I was having issues when serving partial requests (using Range headers) from [whitenoise](https://github.com/evansd/whitenoise) wrapped in WsgiToAsgi. Some HTTP and ASGI servers in front of that complained about extra content in the response body (I've had the issue with traefik in front of uvicorn and with [uvicorn itself when using the httptools protocol handler](https://github.com/encode/uvicorn/blob/master/uvicorn/protocols/http/httptools_impl.py#L517)).

One contributing factor is whitenoise which returns the file handle to the whole file in a [wsgiref.FileWrapper](https://github.com/python/cpython/blob/master/Lib/wsgiref/util.py#L11), seeked to the correct position for the range request. It then expects the server to only read as many bytes as defined in the Content-Length header.

The [WSGI specs allows for this behavior](https://www.python.org/dev/peps/pep-3333/#handling-the-content-length-header) and [gunicorn implements it](https://github.com/benoitc/gunicorn/blob/master/gunicorn/http/wsgi.py#L333) and stops reading after as many bytes as specified in Content-Length.

The [ASGI spec](https://asgi.readthedocs.io/en/latest/specs/www.html#http) does not say much about the consistency between Content-Length header and the "http.response.body" events. The relevant parts could be :
> Protocol servers must flush any data passed to them into the send buffer before returning from a send call.

> Yielding content from the WSGI application maps to sending http.response.body messages.

The first one suggests that any piece of body from an "http.response.body" event must be sent, the other one could mean that all behavior defined for WSGI should apply to ASGI, including not-sending extra body, maybe ?

[uvicorn](https://github.com/encode/uvicorn/blob/master/uvicorn/protocols/http/h11_impl.py#L475), [daphe](https://github.com/django/daphne/blob/master/daphne/http_protocol.py#L255) and [hypercorn](https://gitlab.com/pgjones/hypercorn/-/blob/master/src/hypercorn/protocol/http_stream.py#L147) lean more towards the first interpretation and pass through the whole readable content or simply raise an exception.

If you agree that the WsgiToAsgi adapter should clamp the body I can work on a patch.

Aside from this, the ASGI spec could probably be improved by defining this point. I don't really know if ASGI servers should prefer clamping or forwarding extra body responses but I have the vague feeling that allowing ASGI events to generate incorrectly framed HTTP messages could generate hard to diagnose issues in the future without providing any benefit.

If the sendfile extension from #184 is added, the same point will probably arise, it would be very inconvenient to only be able to use sendfile when serving whole files. WSGI also specifies the clamping behavior [when using sendfile](https://www.python.org/dev/peps/pep-3333/#optional-platform-specific-file-handling) and [gunicorn](https://github.com/benoitc/gunicorn/blob/master/gunicorn/http/wsgi.py#L365) implements it.

kmichel:
This fixes the bug mentioned in #195 where a WSGI server should not send more
output than Content-Length, if the header is present.

After reading more of the spec, this also adds the specified behavior of stopping
the response iteration as soon as we have enough bytes.

The WSGI spec also says that a response body that is too short should also log
or raise an error and close the connection.

This part seems hard to do correctly while also handling HEAD requests:
these requests can contain a Content-Length header and, depending on who you
ask, the WSGI application should or should not return the response body and
expect the WSGI server to remove it. I was not familiar with the issue but it seems
there's no clear consensus so I've opted to leave it alone :)

Since write() is not present in WsgiToAsgi, I also did not try to raise an error if write()
is called again after the response reaches Content-Length.

### DIFF
--- a/asgiref/wsgi.py
+++ b/asgiref/wsgi.py
@@ -29,6 +29,7 @@ class WsgiToAsgiInstance:
     def __init__(self, wsgi_application):
         self.wsgi_application = wsgi_application
         self.response_started = False
+        self.response_content_length = None
 
     async def __call__(self, scope, receive, send):
         if scope["type"] != "http":
@@ -114,6 +115,11 @@ class WsgiToAsgiInstance:
             (name.lower().encode("ascii"), value.encode("ascii"))
             for name, value in response_headers
         ]
+        # Extract content-length
+        self.response_content_length = None
+        for name, value in response_headers:
+            if name.lower() == "content-length":
+                self.response_content_length = int(value)
         # Build and send response start message.
         self.response_start = {
             "type": "http.response.start",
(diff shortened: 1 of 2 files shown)


andrewgodwin (approved):
This looks great - if you rebase, I should have fixed the linting error that snuck onto the main branch, and we can make sure this is all green to go!

andrewgodwin:
Thanks so much for the great PR!

PR #199, opened 2020-10-05
[RFC] add lifespan.shutdown.notice event

justindujardin:
Hi 👋  and thanks for the excellent specification! I would like to propose a change to allow looping endpoints to cleanly exit when an ASGI application shuts down.

### Use Case
An ASGI application exposes an indefinitely looping endpoint using [Server-Sent Events](https://www.w3.org/TR/eventsource/) to push unidirectional data to browser clients without client polling or WebSockets.

### Example 
A counter event-stream sends incrementing numbers to a connected client until they disconnect, or the server shuts down.

```python
import asyncio

state = {"should_exit": False, "counter": 0}


async def is_disconnected(receive) -> bool:
    """Check for a "http.disconnect" message"""
    receive_task = asyncio.create_task(receive())
    try:
        await asyncio.wait_for(receive_task, 1)
        message = receive_task.result()
        if message["type"] == "http.disconnect":
            print("disconnect received")
            return True
    except asyncio.TimeoutError:
        receive_task.cancel()
    return False


async def lifespan(scope, receive, send):
    """Handle lifespan events"""
    while True:
        message = await receive()
        if message["type"] == "lifespan.startup":
            print("startup received")
            await send({"type": "lifespan.startup.complete"})
        elif message["type"] == "lifespan.shutdown.notice":
            state["should_exit"] = True
            print("shutdown notice received")
        elif message["type"] == "lifespan.shutdown":
            print("shutdown received")
            await send({"type": "lifespan.shutdown.complete"})
            return


async def http(scope, receive, send):
    message = await receive()
    assert message["type"] == "http.request"

    while message.get("more_body"):
        message = await receive()

    await send(
        {
            "type": "http.response.start",
            "status": 200,
            "headers": [
                (b"cache-control", b"no-cache"),
                (b"content-type", b"text/event-stream"),
                (b"connection", b"keep-alive"),
            ],
        }
    )

    while True:
        await send(
            {
                "type": "http.response.body",
                "body": f"data: {state['counter']}\n\n".encode("utf-8"),
                "more_body": True,
            }
        )
        state["counter"] += 1
        # Check to see if the client disconnected
        if await is_disconnected(receive):
            break
        else:
            print(f"counter = {state['counter']}")

        # The lifespan.shutdown.notice event fired
        if state["should_exit"]:
            print("shutdown notice received, exiting loop")
            await send(
                {
                    "type": "http.response.body",
                    "more_body": False,
                }
            )
            break

        await asyncio.sleep(3)


    if scope["type"] == "lifespan":
        await lifespan(scope, receive, send)
    elif scope["type"] == "http":
        await http(scope, receive, send)

```

### Problem
With version 2.0 of the lifespan extension, this app can't terminate gracefully without going outside the spec, because the `lifespan.shutdown` event fires "when the server has stopped accepting connections **and closed all active connections**." 

In most cases, firing a shutdown event after all connections have been closed is fine because requests complete and return. However, when you have looping applications that need to break on shutdown, they can't add shutdown handlers to trigger the breakout because they won't fire until the connection has exited.

### Proposal
I explored the minimal changes to support this use-case. Creating a new event looks like the lowest impact way of adding support for signaling connections to gracefully exit. It limits the breakage in existing ASGI frameworks by leaving the "and closed all active connections" requirement for firing `shutdown` while allowing for looping endpoints to exit during application shutdown.

### DIFF
--- a/specs/lifespan.rst
+++ b/specs/lifespan.rst
@@ -88,11 +88,11 @@ Keys:
 
 
 Shutdown Notice - ``receive`` event
-''''''''''''''''''''''''''''''''
+'''''''''''''''''''''''''''''''''''
 
-Sent to the application immediately before the server stops accepting new 
+Sent to the application immediately before the server stops accepting new
 connections and starts shutting down. Useful for signaling long-running or
-looping connections to terminate.
+looping connections, like those using Server-Sent Events, to terminate.
 
 Keys:
 
--- a/specs/lifespan.rst
+++ b/specs/lifespan.rst
@@ -2,7 +2,7 @@
 Lifespan Protocol
 =================
 
-**Version**: 2.0 (2019-03-20)
+**Version**: ?.? (2020-??-??)
 
 The Lifespan ASGI sub-specification outlines how to communicate
 lifespan events such as startup and shutdown within ASGI. This refers to the
@@ -87,6 +87,18 @@ Keys:
 * ``message`` (*Unicode string*) -- Optional; if missing defaults to ``""``.
 
 
+Shutdown Notice - ``receive`` event
+''''''''''''''''''''''''''''''''
+
+Sent to the application immediately before the server stops accepting new 
+connections and starts shutting down. Useful for signaling long-running or
+looping connections to terminate.
+
+Keys:
+
+* ``type`` (*Unicode string*) -- ``"lifespan.shutdown.notice"``.
+
+
 Shutdown - ``receive`` event
 ''''''''''''''''''''''''''''
 
@@ -124,6 +136,8 @@ Keys:
 Version History
 ===============
 
+* ?.? (2020-??-??): Added shutdown.notice for signaling
+  looping applications to exit.
 * 2.0 (2019-03-04): Added startup.failed and shutdown.failed,
   clarified exception handling during startup phase.
 * 1.0 (2018-09-06): Updated ASGI spec with a lifespan protocol.
--- a/specs/lifespan.rst
+++ b/specs/lifespan.rst
@@ -90,9 +90,9 @@ Keys:
 Shutdown Notice - ``receive`` event
 '''''''''''''''''''''''''''''''''''
 
-Sent to the application immediately before the server stops accepting new
-connections and starts shutting down. Useful for signaling long-running or
-looping connections, like those using Server-Sent Events, to terminate.
+Sent to the application immediately after the server stops accepting new
+connections, before it starts shutting down. Useful for signaling long-running
+or looping connections, like those using Server-Sent Events, to terminate.
 
 Keys:
 


tiangolo (approved):
Nice! :nerd_face: This LGTM :rocket:

...although we work together and @justindujardin already knew this LGTM. :sweat_smile: 

---

More details from my side:

I also tried to find a way to handle this with the current spec and I couldn't find anything.

The options I considered:

* The HTTP disconnect `receive` event is sent when the client disconnects, but I think it wouldn't apply when the server is the one disconnecting.
* The Lifespan `receive` shutdown event, as Justin says, is sent after the connections are done, so, Server-Sent Events can't receive it as they are the ones keeping a connection open.

Also, if the `shutdown` event was received before the connections are closed, any resources (e.g. DB connections) could be closed by the `shutdown` event handler, possibly leaving any ongoing connections without the resources they needed.

---

I tested the example app above with the 3 major servers, Uvicorn, Hypercorn, Daphne, and they all behaved similarly (with small differences), none of them had a way to tell the running app that it should close the server connection.

Running the app above, starting a connection (opening a Chrome browser tab, that listens on and receives the data), and then stopping the server with `Ctrl+c`:

* **Uvicorn**: kept waiting until a second `Ctrl+c`, then it terminated the app without sending the `shutdown` (after the forced kill with the second `Ctrl+c`).
* **Hypercorn**: killed the app after a timeout, **and then** sent the `shutdown` event.
* **Daphne**: killed the app almost immediately, I think without a timeout. And didn't send the `shutdown` event.

But the main point is, none of them sent any additional ASGI event until the app keeping the connection and sending the SSEs closed the connection/stopped, like when the tab is closed, or a timeout was reached.

I guess it applies mainly for Server-Sent Events, other streaming responses (e.g. an infinite video stream), or WebSockets.

So it seems to me that a pre `shtudown` event (what Justin is proposing) is needed for these cases.

tomchristie:
> With version 2.0 of the lifespan extension, this app can't terminate gracefully without going outside the spec, because the lifespan.shutdown event fires "when the server has stopped accepting connections and closed all active connections."

Rather than add a new message is it worth considering changing the semantics of the shutdown event slightly, so that it fires once the server has stopped accepting connections?

tiangolo:
> Rather than add a new message is it worth considering changing the semantics of the shutdown event slightly, so that it fires once the server has stopped accepting connections?

I thought of that first, but I think that would make any event handler that closes DB connections terminate them while the ongoing app might still need them.

So, the streaming or SSE response could lose the DB connection it needs before it gets the chance to terminate the response.

I didn't find a straightforward way to circumvent that, but I could be missing something...

justindujardin:
> Rather than add a new message is it worth considering changing the semantics of the shutdown event slightly, so that it fires once the server has stopped accepting connections?

Both @tiangolo and myself initially tried this approach. As he points out there are cases where that wouldn't work out.

Thinking generally about the lifecycle of a looping async program, it always needs an async way of doing cleanup. The same is true for ASGI servers and applications. Because of that, I think it makes sense to provide opportunities for both pieces to do async cleanup when exiting:
 - the application gets "lifespan.shutdown.notice" and has the time between that event and when it returns to do some async thing (as guaranteed by the "and closed all active connections." wording for firing "lifespan.shutdown")
 - the server gets the "lifespan.shutdown" event and has the time between then and when it sends "lifespan.shutdown.complete" 

I could see an argument that "lifespan.shutdown.notice" should send a complete event rather than relying on the prerequisite language for firing "lifespan.shutdown", but I thought keeping it simple would be best.

pgjones:
> The HTTP disconnect receive event is sent when the client disconnects, but I think it wouldn't apply when the server is the one disconnecting.

In my view a http disconnect receive event must be sent when the connection has closed. Including when the server closes it (when the client disconnects). For clarity I think it should be sent when the server is the one disconnecting.

Rather than being a lifespan event, have you considered introducing a http/websocket event (e.g. `http.goaway`) that indicates to the application that the connection will be closed? (I'd imagine this would make it easier for the application to manage state as well). For clarity I expect the application to also receive a `http.disconnect` event when the connection has been closed as per the current spec.

I'm not sure I know what an application could do with an event like this (a pre-disconnect message)? In your example the application seems to send a pseudo goaway message back to the server. However the server is already best placed to send goaway frames/messages to the client without any need for the application to do anything.

tiangolo:
Thanks for the input @pgjones !

I think the main problem here is that in this case, the application is the one that is maintaining a connection open with a streaming response (say, with Server-Sent Events).

> Rather than being a lifespan event, have you considered introducing a http/websocket event (e.g. `http.goaway`) that indicates to the application that the connection will be closed?

So, in the case of Uvicorn, it awaits for the current connections to finish sending their responses before closing them. That's a little difference with Hypercorn in behavior, but it means that if the server is sending an infinite streaming response (as are Server-Sent Events), the response is never "done", and the application ends up keeping the connection alive, so the server never really closes the connection, it doesn't receive new connections but it lets the app finish the existing ones.

Then, in all the cases, Hypercorn after a timeout, Daphne after a timeout (or no timeout?), and Uvicorn after a second `Ctrl+c`, the server forcefully kills the connection, regardless of the internal app state at that point, and regardless of if it was able to clean up properly before finishing or not (e.g. saving state for that request's user in a DB before finishing the streaming response).

But yeah, independent of the semantics, a new HTTP event would probably work too for this specific use case, if that's preferable for the spec.

---

Although, I would actually think that other future protocols that would also be covered by the lifespan of the event loop (apart from HTTP and WebSockets) could benefit from knowing when they have to finish their infinite responses with the `shutdown.notice` before the actual `shutdown` event is received and they are forcefully killed.

For example, imagine some new ASGI spec to send an infinite stream of video, it could send infinite contiguous video segments (like files), in an infinite stream. Right before closing the connection, when receiving the `shutdown.notice` event, this app could make sure it sends the last chunks of bytes to make sure it sent a complete file, instead of leaving the last one incomplete and corrupt. And then after that, the app could finish the stream/tell the server to close the connection (on the app side, not on the client nor server side). And then, the `shutdown` event would be received, and the app could close DB connections, etc. in the `shutdown` handlers.

Also, if it was all done only in the `shutdown` handler, the DB connection could be closed before the last chunks of a video file were sent.

That's why I think it would be nice for the event to be part of the lifespan.

---

Maybe another option for all this could be only to change a little the meaning of the `shutdown` event, and send it before the connections are closed. So, the app could still send some `http` events even after receiving the `shutdown` event. And in this case, it would be the responsibility of the app to coordinate what it has to do before running the rest of the `shutdown` event handlers.

So, it would:

* Receive the `shutdown` event in a lifespan handler
* Store some global state marking the app as "shutting down", that lets the rest of the app read it
* Then the code handling the response (probably in another function) that is running concurrently, would keep consulting that global app state to know if it has to stop, and stop its infinite streaming response, clean up, etc
* Then that same response code would have to indicate that it is done with its cleanup, again in some global app state
* Next, the initial lifespan handler that would have been running concurrently all this time, would have been waiting for this last cleanup flag, and when it's done, it would finally be able to continue with the normal shutdown process, close the DB, etc.

The issue I see from this last option, apart from being quite more convoluted, is that it would end up leaving all the problem to the app/framework implementation, instead of taking advantage of the approach of "turtles all the way down" in the spec, so, e.g. a generic ASGI middleware would not be compatible with these interactions, it would have to be very app/framework specific.

justindujardin:
>In your example the application seems to send a pseudo goaway message back to the server. However the server is already best placed to send goaway frames/messages to the client without any need for the application to do anything.

The example was intended to run against a forked version of Uvicorn that already [has support for this change](https://github.com/encode/uvicorn/pull/800). @tiangolo and I mostly used it to test how existing ASGI frameworks respond in the reproducible dead-lock case, and that this change allows connections to exit gracefully (without force terminating.) It doesn't have much to do with client disconnects, just applications that hang during shutdown.

> I'm not sure I know what an application could do with an event like this (a pre-disconnect message)? 

The runnable example probably lacks clarity, so consider some pseudocode for an [SSE](https://www.w3.org/TR/eventsource/) chat stream:

```python
from typing import List, Optional
from datetime import datetime


def get_new_chats_from_database(last_push: Optional[datetime]) -> List[str]:
    return []  # your code


def should_exit() -> bool:
    return False  # should the loop exit?


async def chat_stream(scope, receive, send):
    state_last_push_time = None
    await send(
        {
            "type": "http.response.start",
            "status": 200,
            "headers": [
                (b"cache-control", b"no-cache"),
                (b"content-type", b"text/event-stream"),
                (b"connection", b"keep-alive"),
            ],
        }
    )

    while True:
        new_messages: List[str] = get_new_chats_from_database(state_last_push_time)
        if len(new_messages) > 0:
            state_last_push_time = datetime.now()
            for msg in new_messages:
                await send(
                    {
                        "type": "http.response.body",
                        "body": f"data: {msg}\n\n".encode("utf-8"),
                        "more_body": True,
                    }
                )
        await asyncio.sleep(3)
        if should_exit():
            print("[SSE] Terminating push connection at server request")
            break
    await send(
        {
            "type": "http.response.body",
            "more_body": False,
        }
    )
    print("[SSE] Stream Complete")
```

When the server shuts down with connected clients, the app needs a signal to know it should break (close its connection) so that the ASGI server can fire the "lifespan.shutdown" event. The addition of the "lifespan.shutdown.notice" event provides a callback that can cause `should_exit` to return true.

---

The particular event for signaling that the applications should exit is not of great concern to us. That said, providing some signal that applications should "begin exiting" is appropriate and consistent with reasonable async programming practices. 

For example, the asyncio event-loop has [multiple](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.stop) [signaling mechanisms](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.shutdown_asyncgens) to cause it to exit. Posix threads [get signaled to exit](https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_exit.html), then [wait for it to happen](https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_join.html). .NET Background workers [send cancel-signals](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.backgroundworker.cancelasync?view=netcore-3.1) to indicate that workers should shut down. Pretty much anywhere you have async code execution, there's an API for signaling that it's time to quit.

The point with all the examples is to highlight that this change (or some variation) is not superfluous, but rather provides a common async cancellation technique that is missing for the Application side of the ASGI spec because it addresses an uncommon use-case. We're happy to adjust this PR to fit the project stakeholders' requirements. Thanks for all the feedback so far ❤️

andrewgodwin:
> For clarity I expect the application to also receive a http.disconnect event when the connection has been closed as per the current spec.

Yes, this is my interpretation of it as well. If any server is not firing `http.disconnect` when it's forcibly closing connections, I'd say that's a bug and should be fixed.

I see the idea here, but the main reason I might object is that any application should _always_ be built so that a sudden exit is not catastrophic; relying on exit events is something I generally encourage my team not to do for this reason ("crash-proof software"). Is there a specific example - maybe the deadlock case you alluded to above - where a server just exiting and force-closing all connections is not good enough?

Obviously if you're doing SSE/websockets, then you already have to deal with people going into tunnels or otherwise losing internet, so that presumably is quite resilient to connections being dropped already, but I want to know what the specific gain would be here that couldn't otherwise be achieved.

justindujardin:
> the main reason I might object is that any application should always be built so that a sudden exit is not catastrophic; relying on exit events is something I generally encourage my team not to do for this reason ("crash-proof software"). 

Thanks for clarifying your expectations. To be clear, catastrophic exits are fine for us, we're only concerned about having to force an exit to get the process to terminate during normal shutdown (e.g. during software update restarts)

> Is there a specific example - maybe the deadlock case you alluded to above - where a server just exiting and force-closing all connections is not good enough?

In our case the ASGI framework (Uvicorn) does not force connections to close when it receives a shutdown signal, but waits for them to exit before continuing to shutdown, which leads to deadlocks during normal shutdown when an SSE app, 
 like the chat example, is running.

Is it possible that the spec needs to be more explicit about how a server should handle terminating existing connections? If one server framework decides to force them closed and another waits patiently for them to exit, that's not desirable, right? 

The only reference I found for the shutdown lifecycle is `lifespan.shutdown` which says `Sent to the application when the server has stopped accepting connections and closed all active connections.` but doesn't specify how the connections should be closed. Is the shutdown behavior addressed somewhere else? 🤔

andrewgodwin:
Ah, great, now I get what you're saying. My expectation is that signalling the server to shutdown should be done using the standard OS signals (e.g. SIGTERM and SIGKILL on Unix-style systems) - if we wanted to carve out a way for a server to be forced to exit from inside itself, I would say that raising `SystemExit` should do it. Nothing in the `lifespan` extension was ever intended to be a signalling mechanism to trigger shutdown, just a way for the server to tell the application when to start any background cleanup loops and then tidy them up.

justindujardin:
>My expectation is that signalling the server to shutdown should be done using the standard OS signals (e.g. SIGTERM and SIGKILL on Unix-style systems) 

@andrewgodwin cool, that's what we do, either directly via ctrl-c in development mode, or indirectly by way of `systemctl restart the_service`. The trouble is that in response to the normal OS signals, Uvicorn does not force close connections and creates a deadlock, while the other popular ASGI frameworks have different behaviors for the same (very reasonable, if uncommon) use-case.

Not to put you on the spot, but if you don't plan to accept this change, can you weigh in on how Uvicorn (and others) should handle shutting down active connections when they receive OS signals to exit? Without your input it's not clear whether I should try to fix a "bug" in Uvicorn, or a "bug" in the spec. 😓

andrewgodwin:
Well, I imagine that `uvicorn` is waiting for connections to close when it's getting SIGTERM, but since SIGKILL is untrappable, I presume that would still work. Waiting for existing connections to exit on SIGTERM seems sensible to me, honestly.

I do see what you mean - given you have infinitely long-lived connections, SIGTERM is useless to you and so you can only use SIGKILL - but the way I'd handle this is by... just killing the whole process :) I know that's not very useful, so I'd suggest that the behaviour I might expect/want to see in uvicorn is to wait 60 seconds for existing coroutines to exit and then force them shut.

andrewgodwin:
...to elaborate a bit further, I totally see how a `lifespan.shutdown.notice` event would appear to help, but I don't like adding things into systems that are not guaranteed to be called.

If we add a shutdown-notice event, it is never going to be sent 100% of the time - far less than that in any stable system, I'd wager - and so building anything that actually relies on it is going to be worse than nothing, since silently not running code is never good.

I'd much rather encourage the idea in the spec that a coroutine/application can be killed at any time, for any reason, and the way to build resilient software is to just build that in as a design consideration.

tomchristie:
Pile in if I'm getting this wrong, but I *think* the missing part in the puzzle here is that the *application side* needs to be listening for disconnect events, right?

Regular ol' request/response handlers can just ignore that, and always run to completion, but SSE endpoints need to be listening for `"http.disconnect"`.

There is *perhaps* one aspect of the spec here that could do with thinking about, which is that I think *ideally* server implementations ought to send `"http.disconnect"` on all HTTP connections as soon as server-shutdown is initiated, but *MAY* still want to allow a grace period before they then force-close the task, in order to still allow regular ol' HTTP request/response cycles to run to completion.

In the case where the app is serving SSE or long-polling connections, but failing to listen for the disconnect event, then the server should end up force-closing them at the end of the grace period.

tomchristie:
(I don't recall right now exactly what Uvicorn does, but the above seems to me to be generally what *ought* to happen here)

andrewgodwin:
Right, I would imagine that a sensible graceful shutdown is:

- Stop accepting new connections
- Wait [30s] for coroutines to complete and close themselves out
- Forcibly disconnect the rest, sending them `http.disconnect` in the process (as this should get sent no matter how the connection is terminated)
- Give them another [10s] to finish and tidy up
- Force kill any remaining coroutines and exit

tomchristie:
Interesting yeah, that makes sense too, tho it's *slightly* different from what I was thinking, which is:

* Stop accepting new connections, sending them http.disconnect in the process
* Wait [30s] for coroutines to complete and close themselves out (still read any response messages in the period)
* Force kill any remaining coroutines and exit

pgjones:
@tomchristie Hypercorn should be doing what you describe at the moment but I think the double timeout @andrewgodwin suggests is better so I'll likely change.

justindujardin:
> I think ideally server implementations ought to send "http.disconnect" on all HTTP connections as soon as server-shutdown is initiated, but MAY still want to allow a grace period before they then force-close the task

@tomchristie This change would completely accommodate our use-case. Thanks for taking the time to understand the problem 💯 🙇 

> - Stop accepting new connections
> -  Wait [30s] for coroutines to complete and close themselves out
> -  Forcibly disconnect the rest, sending them http.disconnect in the process (as this should get sent no matter how the connection is terminated)

> Hypercorn should be doing what you describe at the moment but I think the double timeout @andrewgodwin suggests is better so I'll likely change.

@pgjones @andrewgodwin I think we're still not fully understanding each other because the new behaviors you suggest would impose a definite wait period on my application before it could exit, which is not what I think most people would consider graceful.

The important difference is firing an event that the application can handle before entering a waiting period, as Tom suggested. When you wait before telling the application anything, it privileges the typical use-case where applications naturally exit on their own, and requires loops like mine to wait for the duration of the timeout because they have no way of knowing that they should return. 

I suppose if Uvicorn has desirable behavior I shouldn't care about other frameworks, but I think that if you consider it deeply, you'll find that Tom's suggestion makes the most sense as an alternative to what I've proposed here.

PR #201, opened 2020-10-11
Reduce CPU usage in CurrentThreadExecutor.run_until_future()

asedeno:
https://github.com/django/asgiref/blob/7dba5ff61313dfd1535de200590b9155abde20f0/asgiref/current_thread_executor.py#L56-L68

`CurrentThreadExecutor.run_until_future()` keeps busy checking two things:

1. Do I have something to do?
2. Is the future I'm tracking done?

with a short sleep in between iterations.

I noticed this when I was wondering why some threads I had running, which should be doing basically nothing, were using noticeable, though relatively low, CPU. (~5%) This was in three long-lived django helper processes running async event loops in an app I'm working on where that process count will scale as 2+n_users, so I had concerns about this CPU use as I prepared to add more users.

Because we're checking two things, we can't block on checking one thing, necessitating the sleep to prevent hot spinning. But is this really necessary?

In this PR, I'm proposing that it's not. The `submit()` method on the `CurrentThreadExecutor` pushes a `_WorkItem` onto the `CurrentThreadExecutor`'s `_work_queue`. If we could put a sentinel value on the queue when the future was finished, then we could block on `_work_queue.get()` and not waste cycles checking on the future itself. `future.add_done_callback()` allows us to register a function as a callback on the future, and it will be called with the future itself as an argument when it's done. `_work_queue.put()` takes an argument and puts it on the queue. Put the two together, and we get exactly this.

Adding `future.add_done_callback(self._work_queue.put)` gives us our sentinel value (`future`) at exactly the right time. Now `self._work_queue.get()` can be called in blocking fashion, we'll either get a job or an indicator that we're done, and the sleep can be dropped entirely.

### DIFF
--- a/asgiref/current_thread_executor.py
+++ b/asgiref/current_thread_executor.py
@@ -1,6 +1,5 @@
 import queue
 import threading
-import time
 from concurrent.futures import Executor, Future
 
 
@@ -51,21 +50,17 @@ class CurrentThreadExecutor(Executor):
             raise RuntimeError(
                 "You cannot run CurrentThreadExecutor from a different thread"
             )
-        # Keep getting work items and checking the future
+        future.add_done_callback(self._work_queue.put)
+        # Keep getting and running work items until we get the future we're waiting for
+        # back via the future's done callback.
         try:
             while True:
                 # Get a work item and run it
-                try:
-                    work_item = self._work_queue.get(block=False)
-                except queue.Empty:
-                    # See if the future is done (we only exit if the work queue is empty)
-                    if future.done():
-                        return
-                    # Prevent hot-looping on nothing
-                    time.sleep(0.001)
-                else:
-                    work_item.run()
-                    del work_item
+                work_item = self._work_queue.get()
+                if work_item is future:
+                    return
+                work_item.run()
+                del work_item
         finally:
             self._broken = True
 


asedeno:
I just found `Future.add_done_callback()` and with that, I now have a much smaller change that I am also more confident in.

asedeno:
Also, an alternative that's cute, but probably unnecessarily clever:
Skip the new `_stop()` method altogether, make the done callback `self._work_queue.put`, and use the future itself as the sentinel value. It brings new meaning to `run_until_future()`. 😄

asedeno:
I've been running this for myself now for over a week, no issues, and no ~5% CPU usage for what should be an idle process. @andrewgodwin, any thoughts on this, either this version of the alternative I mentioned about a week ago?

andrewgodwin:
I'll get to this when I get a chance, but it might be a month or so from now - sorry. Life is getting in the way.

asedeno:
Life has a tendency to do that. Good luck!

andrewgodwin (approved):
OK, I like the look of this, and especially the fact that it's a net code deletion. Ready to get it merged in?

asedeno:
As far as the code goes, yes. Give me 5 minutes and I'll rebase it onto master and adjust the commit message to reflect the latest commit.

asedeno:
Ready when you are.

andrewgodwin:
Thanks for your work on this!

PR #220, opened 2020-12-24
Copy unix socket usage to Websocket scope

pgjones:
This was missed in e14932bcc6c0287898ae2ba24d1f872a2c5aedc5 but
applies.


andrewgodwin:
Ta!

PR #221, opened 2020-12-24
Add typing for ASGI Scopes, messages, and frameworks

pgjones:
These are useful for users of asgiref to type check their ASGI
usage. It also serves as the reference typing for the ASGI
Specification.

### DIFF
--- a/asgiref/typing.py
+++ b/asgiref/typing.py
@@ -1,0 +1,204 @@
+from typing import Awaitable, Callable, Dict, Iterable, Optional, Tuple, Type, Union
+
+try:
+    from typing import Literal, Protocol, TypedDict
+except ImportError:
+    from typing_extensions import Literal, Protocol, TypedDict  # type: ignore
+
+
+class ASGIVersions(TypedDict):
+    spec_version: str
+    version: Union[Literal["2.0"], Literal["3.0"]]
+
+
+class HTTPScope(TypedDict):
+    type: Literal["http"]
+    asgi: ASGIVersions
+    http_version: str
+    method: str
+    scheme: str
+    path: str
+    raw_path: bytes
+    query_string: bytes
+    root_path: str
+    headers: Iterable[Tuple[bytes, bytes]]
+    client: Optional[Tuple[str, int]]
+    server: Optional[Tuple[str, Optional[int]]]
+    extensions: Dict[str, dict]
+
+
+class WebsocketScope(TypedDict):
+    type: Literal["websocket"]
+    asgi: ASGIVersions
+    http_version: str
+    scheme: str
+    path: str
+    raw_path: bytes
+    query_string: bytes
+    root_path: str
+    headers: Iterable[Tuple[bytes, bytes]]
+    client: Optional[Tuple[str, int]]
+    server: Optional[Tuple[str, Optional[int]]]
+    subprotocols: Iterable[str]
+    extensions: Dict[str, dict]
+
+
+class LifespanScope(TypedDict):
+    type: Literal["lifespan"]
+    asgi: ASGIVersions
+
+
+WWWScope = Union[HTTPScope, WebsocketScope]
+Scope = Union[HTTPScope, WebsocketScope, LifespanScope]
+
+
+class HTTPRequestEvent(TypedDict):
+    type: Literal["http.request"]
+    body: bytes
+    more_body: bool
+
+
+class HTTPResponseStartEvent(TypedDict):
+    type: Literal["http.response.start"]
+    status: int
+    headers: Iterable[Tuple[bytes, bytes]]
+
+
+class HTTPResponseBodyEvent(TypedDict):
+    type: Literal["http.response.body"]
+    body: bytes
+    more_body: bool
+
+
+class HTTPServerPushEvent(TypedDict):
+    type: Literal["http.response.push"]
+    path: str
+    headers: Iterable[Tuple[bytes, bytes]]
+
+
+class HTTPDisconnectEvent(TypedDict):
+    type: Literal["http.disconnect"]
+
+
+class WebsocketConnectEvent(TypedDict):
+    type: Literal["websocket.connect"]
+
+
+class WebsocketAcceptEvent(TypedDict):
+    type: Literal["websocket.accept"]
+    subprotocol: Optional[str]
+    headers: Iterable[Tuple[bytes, bytes]]
+
+
+class WebsocketReceiveEvent(TypedDict):
+    type: Literal["websocket.receive"]
+    bytes: Optional[bytes]
+    text: Optional[str]
+
+
+class WebsocketSendEvent(TypedDict):
+    type: Literal["websocket.send"]
+    bytes: Optional[bytes]
+    text: Optional[str]
+
+
+class WebsocketResponseStartEvent(TypedDict):
+    type: Literal["websocket.http.response.start"]
+    status: int
+    headers: Iterable[Tuple[bytes, bytes]]
+
+
+class WebsocketResponseBodyEvent(TypedDict):
+    type: Literal["websocket.http.response.body"]
+    body: bytes
+    more_body: bool
+
+
+class WebsocketDisconnectEvent(TypedDict):
+    type: Literal["websocket.disconnect"]
+    code: int
+
+
+class WebsocketCloseEvent(TypedDict):
+    type: Literal["websocket.close"]
+    code: int
+
+
+class LifespanStartupEvent(TypedDict):
+    type: Literal["lifespan.startup"]
+
+
+class LifespanShutdownEvent(TypedDict):
+    type: Literal["lifespan.shutdown"]
+
+
+class LifespanStartupCompleteEvent(TypedDict):
+    type: Literal["lifespan.startup.complete"]
+
+
+class LifespanStartupFailedEvent(TypedDict):
+    type: Literal["lifespan.startup.failed"]
+    message: str
+
+
+class LifespanShutdownCompleteEvent(TypedDict):
+    type: Literal["lifespan.shutdown.complete"]
+
+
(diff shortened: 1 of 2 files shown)


pgjones:
These are from Hypercorn (so I think they are correct, or Hypercorn is wrong :o).

euri10:
This is dope, thanks !

andrewgodwin:
You are a wonderful person for doing this, I was just thinking about adding typing now we're using it more at work. Get that linter happy and I will smash the merge button.

andrewgodwin:
(and maybe start working these into the actual ASGI code over the xmas break, since it's about type we had a little typing in there alongside tests)

asgiref/typing.py
```diff
@@ -0,0 +1,204 @@
...
+        ...
+
+    async def __call__(
+        self, receive: ASGIReceiveCallable, send: ASGISendCallable
+    ) -> None:
+        ...
+
+
+ASGI2Framework = Type[ASGI2Protocol]
+ASGI3Framework = Callable[
```
florimondmanca, line 196: Thoughts on naming these eg `ASGI3App` instead? Curious about the motivation for "framework" here. Just nitpicking though.

florimondmanca (commented):
_Super_ nice :)

We'll be able to properly use these in Uvicorn by adding a dependency on asgiref then, correct?

florimondmanca:
@pgjones Do we need to add a `py.tyed` containing `partial\n`? I believe that's required for mypy to parse through asgiref when importing it from other packages — or we'd get "Module asgiref is missing" errors or similar. Might also require a tweak to `include_package_data` in `setup.py`? I guess we'd have to check on a dummy test package that type hints would be correctly recognized and used by mypy.

pgjones:
asgiref itself isn't typed, so a `py.typed` at this stage would be misleading (although yes when it is :)). Mypy will be able to understand these typings if they are imported and used.

asgiref/typing.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+class WebsocketDisconnectEvent(TypedDict):
+    type: Literal["websocket.disconnect"]
+    code: int
+
+
+class WebsocketCloseEvent(TypedDict):
+    type: Literal["websocket.close"]
+    code: Optional[int]
```
Kludex, line 124: I think there's a small inconsistency. 

In the [specs](https://asgi.readthedocs.io/en/latest/specs/www.html), specifically in the [HTTP Connection Scope](url), we can see the following:
- `asgi["spec_version"]` (_Unicode string_) – Version of the ASGI HTTP spec this server understands; one of `"2.0"` or `"2.1"`. Optional; if missing assume `2.0`.

But above you can see:
```python
class ASGIVersions(TypedDict):
    spec_version: str
    version: Union[Literal["2.0"], Literal["3.0"]]
```

On the comment line, we have the `WebsocketCloseEvent`, which on the specs says:
- `code` (_int_) – The WebSocket close code, as per the WebSocket spec. Optional; if missing defaults to `1000`.

So following the same line of reasoning, this should be `int`, or the above should be `spec_version` should be `Optional`.

Did I miss something?

tiangolo:
This is awesome! :rocket:

A small note to have in mind, as @Kludex points out, there's probably a caveat with the optionals.

I don't see an obvious way to solve it, as I don't see a standard way to declare these optional types perfectly, so I'll just share the info I have (as I understand it).

---

The issue is that `Optional[SomeType]` actually means "the union of `SomeType` and `None`". But that would imply an explicit `None`, not that the key could be entirely omitted from the dictionary.

And then, using `TypedDict`, it is assumed that all the keys are required: https://docs.python.org/3/library/typing.html#typing.TypedDict

> By default, all keys must be present in a TypedDict. It is possible to override this by specifying totality.

And the example is:

```Python
class point2D(TypedDict, total=False):
    x: int
    y: int
```

But then that means that _any_ of the keys can be omitted, which is also not what is wanted here.

My point is, there is no official way to type a dictionary that could have some specific keys missing. I think the only option here is to assume a convention and take the drawbacks:

1. Declare the dicts using `total=False`, although that makes all the keys possibly missing, which is not great.
2. Mark the optional keys with `Optional[X]`, although that will really mean `Union[X, None]`, and when `None` is not passed explicitly in the dict, linters will (or should/would) complain, which is also not great.

---

I would personally prefer option 2, as that would be annoying when writing code showing false-positives but at least wouldn't accept invalid code that would later break in production. But that's also in my subjective point of view.

---

In any case, I think these types are awesome, thanks for putting the effort to write them. :nerd_face: :rocket: :coffee:

florimondmanca:
There's a 3rd option, which is duplicating the dict, with one version containing the key and the other not, and declaring a Union of those as the final type. Convoluted and unsuitable if there are many non mandatory keys. But maybe... ?

tiangolo:
Ah, very clever @florimondmanca ! :nerd_face: :rocket: 

I think that would be the more formal way to do it indeed. I guess it would get problematic to maintain here in the spec for dicts with several optional parameters, as it would require a lot of types with each of the possible combinations (I think "combinatorial explosion" is the term :thinking:). But yeah, maybe that could work for some cases? I don't know.

pgjones:
I thought about (how to type) omitted keys for Hypercorn, and my view is that there isn't a worthwhile advantage to omitting a key. This is actually the area in which I've had the most ASGI-bugs with Quart/Hypercorn and I'd prefer the specification to state that the keys must be present with a default value.

Practically, using these type hints will result in mypy insisting that the defaults be added when scopes/messages are created, whilst not affecting using messages/scopes. I think this is an acceptable trade off given that trying to type the optionality of the current ASGI spec is quite tricky using the current tooling.

asgiref/typing.py
pgjones: Yep, the specification uses Application - have changed.

florimondmanca (approved):
Stellar. 😄

asgiref/typing.py
Kludex: > I thought about (how to type) omitted keys for Hypercorn, and my view is that there isn't a worthwhile advantage to omitting a key. This is actually the area in which I've had the most ASGI-bugs with Quart/Hypercorn and I'd prefer the specification to state that the keys must be present with a default value.

@pgjones Meaning that we should remove the optional in this case, is that it?

```suggestion
    code: int
```

pgjones: Yep, thanks. Changed.

tiangolo (approved):
This looks awesome! :clap: :rocket:

I just checked all the code while re-reading the spec part by part. Great job! :nerd_face: :coffee:

Kludex (approved):

sm-Fifteen:
Using unions instead of a class hierarchy is also useful to avoid running into python/mypy#7435, since TypedDict also does not support changing the type of a field in a subclass, regardless of compatibility between field types.

> There's a 3rd option, which is duplicating the dict, with one version containing the key and the other not, and declaring a Union of those as the final type. Convoluted and unsuitable if there are many non mandatory keys. But maybe... ? Edit: hmm, maybe not. 😜
> 
> Otherwise agreed with @tiangolo, option 2 (which is almost the current state?) sounds like the most suitable!

@florimondmanca: How about splitting the type to have a total dict with only the required keys, a partial dict with only the optional keys, and declaring the real type as a union of both?

```py
from typing import Awaitable, Callable, Dict, Iterable, Optional, Tuple, Type, Union
from typing import Literal, Protocol, TypedDict

class HTTPRequestEventReq(TypedDict):
	type: Literal["http.request"]
	body: bytes

class HTTPRequestBodyOpt(TypedDict, total=False):
	more_body: bool

class HTTPRequestBody(HTTPRequestEventReq, HTTPRequestBodyOpt):
	pass

good_some: HTTPRequestBody = {
	'type': "http.request",
	'body': b'',
}

good_all: HTTPRequestBody = {
	'type': "http.request",
	'body': b'',
	'more_body': False,
}

bad_missing: HTTPRequestBody = {
	'type': "http.request",
	'more_body': False,
}

bad_extra: HTTPRequestBody = {
	'type': "http.request",
	'body': b'',
	'foo': 'bar',
}
```

andrewgodwin (approved):

andrewgodwin:
I'm happy with this as is; let me know if you want to merge it or if you want to refine the typing a little with the omitted keys handled somehow. I personally think it's OK to not worry about handling omissions and be complete, or we could bring in something else that handles them OK (I know Pydantic does this, but it's not quite standard typing)

pgjones:
I'd prefer to [merge as is](https://github.com/django/asgiref/pull/221#issuecomment-751392284).

andrewgodwin:
Done! Thanks for your work on this, I'm excited to get the actual methods/functions typed now as well.

graingert:
> Mypy will be able to understand these typings if they are imported and used.

mypy will ignore these types unless `py.typed` is present

pgjones:
> mypy will ignore these types unless py.typed is present

I checked this again, and I was wrong mypy completely ignores without the marker (even when directly imported as I thought worked). See #231

pgjones:
[PEP 655](https://www.python.org/dev/peps/pep-0655/) is something we should keep an eye on - it will allow for required (and not required) typed dict keys if accepted.

graingert:
@pgjones it might be better to use `@attr.s` or `@dataclasses.dataclass` objects so defaults can be set cleanly

sm-Fifteen:
> [PEP 655](https://www.python.org/dev/peps/pep-0655/) is something we should keep an eye on - it will allow for required (and not required) typed dict keys if accepted.

@pgjones: Like the PEP highlights and [I pointed out earlier](https://github.com/django/asgiref/pull/221#issuecomment-760273366), you can use inheritance between a total and a non-total typed dict to have both required and non-required keys. It's somewhat cumbersome, but it does work under current versions of Python.

Issue #223, opened 2021-01-08
Django's use of asgiref/sync.py


a-feld:
Hi @andrewgodwin , thanks for all your help on this library and Django! At @newrelic we've been working on adding automated instrumentation support for Django's ASGI mode.

Django internally uses asgiref's `SyncToAsync` and `AsyncToSync` implementations. We've been struggling to figure out what the intended behavior is for Django's use case with `thread_sensitive`.

According to the [docs](https://docs.djangoproject.com/en/3.1/topics/async/) the intent is to have it per-request:
>  ...then Django must use a thread per request to safely emulate a synchronous environment for it.

However, the way the asgiref library is used in django results in the `single_thread_executor` being scoped to the process, meaning that it's 1 thread _total_.

It looks to me like asgiref is intentional in having only 1 thread pool **per process** based on the tests, but I'm thinking Django might be resilient to multi-threading as long as all sync work is scheduled on the same thread (1 thread **per request**).

I have a patch to change over to a request-specific single thread pool that I'm happy to contribute, but I don't want to ship it up without first understanding if this is the intent in both asgiref & Django.

### TL/DR :)
In short:
* Is the intent of Django to run 1 sync thread **per request** or **per process**?
* Is the intent of the `thread_sensitive` parameter to be 1 thread **per process** or **per asyncio root task**?

#### urls.py
```python
import time

from django.http import HttpResponse
from django.urls import path

def view(request):
    print("STARTING")
    time.sleep(10)
    return HttpResponse("*")

urlpatterns = [
    path("", view),
]
```

#### settings.py
```python
ROOT_URLCONF = "urls"
SECRET_KEY = "shhh"
DEBUG = True
```

```bash
daphne asgi:application &
curl localhost:8000 &
curl localhost:8000
```

I'm wondering if with this example app, we expect it to run for 10 seconds or 20 seconds?

andrewgodwin:
The intention of `thread_sensitive` is to run one thread per request - however, this is implemented as one thread per process currently, as that was the safest and most compatible way of achieving this. Obviously, this will be relatively OK with multi-process worker scenarios and not great with multi-threaded ones.

[...]

PR #224, opened 2021-01-11
Add support for context-specific thread pool executors.

a-feld:
As discussed in #223 this PR implements context-specific thread pool capabilities. It does so by creating an input variable (definitely open to updating the name of that input parameter 😛 ) allowing callers of `sync_to_async` to specify a function that returns the current context.

The intended usage of this API is to use [`contextvars` in Django](https://github.com/django/django/compare/master...a-feld:asyncio-perf-improvements) to allow a thread pool per-request.

This PR maintains the existing safety of using a global thread pool per-process by default.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -163,9 +163,14 @@ The first compromise you get to might be that ``thread_sensitive`` code should
 just run in the same thread and not spawn in a sub-thread, fulfilling the first
 restriction, but that immediately runs you into the second restriction.
 
-The only real solution is to essentially have a variant of ThreadPoolExecutor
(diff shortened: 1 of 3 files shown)


andrewgodwin:
Yup, looks good to me - happy to merge this now if you want, or wait until you have a matching Django patch otherwise!

a-feld:
Sounds good, I'm ready to merge if you are! 😄 
I pushed up some docstrings / readme updates to include this parameter as well!

andrewgodwin:
Great, thanks!

PR #226, opened 2021-01-12
Remove typo from invalid character.

jhtimmins:
There was an invalid character in `missing` that showed up as `miss ing`.


andrewgodwin:
Thanks!

PR #227, opened 2021-01-14
Internalize thread sensitive context management in asgiref/sync.py

a-feld:
Rather than using an external context management function, this commit
updates the thread_sensitive mode to reference an internal contextvar.

This approach mainly allows for the thread executor to be shut down when
the context exits, rather than relying on the thread pool shutting down
when going out of scope.

This approach also has other advantages:
* Maintains the existing sync_to_async interfaces.

* Context variable management cannot be broken by omitting the
  current_context_func arg

* This approach opens the door to more complex management techniques
  such as enforcing a thread pool limit.

Django PR showing implementation: https://github.com/django/django/pull/13882

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -163,14 +163,9 @@ The first compromise you get to might be that ``thread_sensitive`` code should
 just run in the same thread and not spawn in a sub-thread, fulfilling the first
 restriction, but that immediately runs you into the second restriction.
 
-By default, a variant of ThreadPoolExecutor executes any ``thread_sensitive``
-code on the outermost synchronous thread - either the main thread, or a single
-spawned subthread.
-
-There is an option to override the default single-threaded behavior so that
-there is 1 synchronous thread per context. If ``current_context_func`` is
-specified, this function will be called to retrieve the current context. There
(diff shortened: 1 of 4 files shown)


a-feld:
Thoughts on this approach @andrewgodwin ? PR is in draft because there are a couple more tests to write. I was racking my brain with how to manage ThreadPoolExecutors with the previous approach - this seems to be a bit more workable 😄 
[Django PR](https://github.com/django/django/pull/13882)

andrewgodwin:
I'll try to take a look at this next weekend; don't have much time right now

andrewgodwin:
I do prefer this approach; shame it won't work on 3.6 but then people should hopefully be mostly off that soon (especially for async).

What else do you want to get done before you consider it ready to land? Looks pretty complete to me.

a-feld:
Awesome! Yeah I think this PR is complete, I'm ready to merge if you are! The only thing remaining on the Django side is some additional testing, then it should be good to go! 🚀

andrewgodwin:
Awesome, thanks for this!

Issue #229, opened 2021-01-20
How should servers utilise `root_path`


pgjones:
`root_path` is often considered equivalent to WSGI's `SCRIPT_NAME`, yet usage of `root_path` by servers differs in that WSGI servers seem to split out the `SCRIPT_NAME` from the `PATH_INFO` whereas ASGI servers (Daphne, Hypercorn, Uvicorn) do not,

I find (for all 3 ASGI servers, with daphne shown),
```
$ SCRIPT_NAME="/bob" gunicorn t110:app
>> environ = {'RAW_URI': '/bob/', 'PATH_INFO': '/', 'SCRIPT_NAME': '/bob', ...}

$ daphne --root-path "/bob" t110:app
>> scope = {'path': '/bob/', 'raw_path': b'/bob/', 'root_path': '/bob', ...}
```
whereas I think the scope should be,
```
>> scope = {'path': '/', 'raw_path': b'/bob/', 'root_path': '/bob', ...}
```

I.e. the server should 404 if the request target does not start with the `root_path`, and if it does remove it from the `path`, but keep it in the `raw_path`. Are the current ASGI servers correct, or does the ASGI spec need to be clearer about this?

davidism:
ASGI documents `root_path` as "same as `SCRIPT_NAME` in WSGI" https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope. That means it should be stripping `root_path` from `path`, otherwise route matching won't work as expected when mounted under a path.

davidism:
This seems to be the relevant bug in Daphne: https://github.com/django/daphne/issues/125. I couldn't find one for Uvicorn.

euri10:
Interesting, we had one user complaining recently it wasn't working as
expected, https://github.com/encode/uvicorn/issues/937

I must admit I rapidly tested against daphne and hypercorn, and as same
routes were served the same way as in uvicorn I maybe diagnosed too fast
this wasn't a problem, but seeing this I'm not sure about it now, not
familiar enough with the SCRIPT_NAME behavior.
I always thought root_path was meant to be used with a reverse proxy

davidism:
`root_path` is also used in reverse proxy situations, although it also needs to be supported by the A/WSGI server.

[...]

PR #230, opened 2021-01-21
Add a note on how root_path should be used

pgjones:
This follows WSGI, as root_path is considered the equivalent to
SCRIPT_NAME and ensures that the root_path can be used as a global
prefix to all routes in the application.

### DIFF
--- a/specs/www.rst
+++ b/specs/www.rst
@@ -414,6 +414,24 @@ with the close code passed in the message (or 1000 if none is specified).
   spec.  Optional; if missing defaults to ``1000``.
 
 
+Path, raw_path and root_path
+----------------------------
+
+The ``raw_path`` is the path component of the request target, which is
+then decoded and unquoted to form an initial path. The ``root_path``
+(defaults to ``""``) prefix is removed from the initial path to create
+the ``path``. If this isn't possible, or the ``path`` does not start
+with a ``/`` the server must respond with a 404 response. For example,
+
+Request target root_path 404? path
+============== ========= ==== ========
+/foo/bar       ""        No   /foo/bar
+/foo/bar       /foo      No   /bar
+/foo/bar       /bar      Yes
+/foo/bar       /foo/bar  Yes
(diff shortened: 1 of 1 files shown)


davidism:
I guess it depends on how you deal with these internally, but the spec probably shouldn't say to base this off `raw_path`. The raw path is optional and may be inconsistent depending on proxies. `root_path` should be taken off the beginning of `path`, the value that's already been decoded.

pgjones:
I've updated to try and make this clearer, and whilst slightly adjusting the rule.

specs/www.rst
```diff
@@ -414,6 +414,24 @@ with the close code passed in the message (or 1000 if none is specified).
...
+(defaults to ``""``) prefix is removed from the initial path to create
+the ``path``. If this isn't possible, or the ``path`` does not start
+with a ``/`` the server must respond with a 404 response. For example,
+
+Request target root_path 404? path
+============== ========= ==== ========
+/foo/bar       ""        No   /foo/bar
+/foo/bar       /foo      No   /bar
+/foo/bar       /bar      Yes
+/foo/bar       /foo/bar  Yes
```
pgjones, line 431: I'm unsure this is the right logic...

davidism: `/foo/bar` should succeed and the path would be `/` (since it would be empty then have a leading slash prepended).

specs/www.rst
```diff
@@ -414,6 +414,24 @@ with the close code passed in the message (or 1000 if none is specified).
...
+/foo/bar/      /foo/bar  No   /
```
davidism, line 432: You should add an example of requesting `/foobar` with `root_path="/foo"` to show that it will 404.

andrewgodwin:
This is not how `path` is specified - see #229. I'm not willing to accept this unless we have all servers in agreement.

pgjones:
Closing, will adapt Quart and WSGI middleware instead.

PR #232, opened 2021-01-23
fix type annotations for when run with mypy on Python 3.6

graingert:


asgiref/typing.py
```diff
@@ -1,9 +1,9 @@
 from typing import Awaitable, Callable, Dict, Iterable, Optional, Tuple, Type, Union
```
graingert, line 1:
```suggestion
import sys
from typing import Awaitable, Callable, Dict, Iterable, Optional, Tuple, Type, Union
```

andrewgodwin:
Thanks!

PR #233, opened 2021-01-24
 type check annotations  (and drop support for python 3.5)

graingert:


asgiref/timeout.py
```diff
@@ -115,14 +113,15 @@ def _cancel_task(self) -> None:
             self._cancelled = True
 
 
-def current_task(loop: asyncio.AbstractEventLoop) -> "asyncio.Task[Any]":
-    if PY_37:
-        task = asyncio.current_task(loop=loop)  # type: ignore
+def current_task(loop: asyncio.AbstractEventLoop) -> "Optional[asyncio.Task[Any]]":
```
graingert, line 116: it might be worth updating the vendoring here: https://github.com/aio-libs/async-timeout/blob/master/async_timeout/__init__.py#L205-L209

andrewgodwin:
👍 Thanks for working on this!

Issue #234, opened 2021-02-02
Missing reason in close event


michallowasrzechonek-silvair:
If I'm not mistaken, the RFC allows applications to terminate websocket connections with optional UTF-8 encoded `reason` field: https://tools.ietf.org/html/rfc6455#section-5.5.1

At the moment doesn't seem supported by ASGI spec: https://asgi.readthedocs.io/en/latest/specs/www.html#close-send-event allows the application to set the `code`, but not the `reason`.

andrewgodwin:
Yes, we chose not to support this because I'd almost never seen it used in the wild, and because the `code` is usually what apps want anyway rather than arbitary bytes. I'm open to adding it if there's sufficient demand.

pgjones:
I've also been asked about this for Quart. Would you accept a PR to add it if I wrote it?

andrewgodwin:
Yup, totally.

michallowasrzechonek-silvair:
Thanks!

PR #235, opened 2021-02-02
Added reason to the WebSocket close

pgjones:
This is allowed as per RFC 6455 but previously not supported in
ASGI. This allows servers/applications to use the reason if desired.

See also #234

### DIFF
--- a/specs/www.rst
+++ b/specs/www.rst
@@ -271,7 +271,7 @@ metadata (mostly from the HTTP request line and headers):
   if missing defaults to ``None``.
 
 * ``query_string`` (*byte string*) -- URL portion after the
-  ``?``. Optional; if missing default is empty string.
+  ``?``. Optional; if missing or ``None`` default is empty string.
 
 * ``root_path`` (*byte string*) -- The root path this application is
   mounted at; same as ``SCRIPT_NAME`` in WSGI. Optional; if missing
--- a/specs/www.rst
+++ b/specs/www.rst
@@ -2,7 +2,7 @@
 HTTP & WebSocket ASGI Message Format
 ====================================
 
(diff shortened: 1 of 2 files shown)


specs/www.rst
```diff
@@ -413,6 +414,9 @@ with the close code passed in the message (or 1000 if none is specified).
 * ``code`` (*int*) -- The WebSocket close code, as per the WebSocket
   spec.  Optional; if missing defaults to ``1000``.
 
+* ``reason`` (*Unicode string*) -- A reason given for the closure, can
+  be any string. Optional; if missing default is empty string.
```
andrewgodwin, line 418: Given the type declared, should this be "if missing or None"?

pgjones: I've gone for the same wording as the [`query_string`](https://github.com/django/asgiref/blame/master/specs/www.rst#L94), should I change both?

andrewgodwin: Yes, let's update both, explicit is better here I think.

pgjones: Done.

pgjones:
(Sorry messed up the branch, now ready)

andrewgodwin:
Thanks for getting this done!

Kludex:
Shouldn't I be able to see 2.3 instead of 2.1 on [this page](https://asgi.readthedocs.io/en/latest/specs/www.html) by now? Am I missing something here?

andrewgodwin:
My bad - looks like type annotations being added killed the build as it was still on Python 2, somehow. Should be all fixed now (we're up to version 3)

Kludex:
Perfect! Thanks @andrewgodwin ! :)

PR #236, opened 2021-02-04
configure pyupgrade via pre-commit

graingert:
also black was missing some files, eg `setup.py` and `docs/conf.py`

I'd recommend configuring https://pre-commit.ci/


setup.py
```diff
@@ -1,3 +1,3 @@
 from setuptools import setup  # type: ignore[import]
 
-setup(name='asgiref')
+setup(name="asgiref")
```
graingert, line 3: @andrewgodwin why is this here? it should work with just
```suggestion
setup()
```

andrewgodwin: Presumably it was a copy-and-paste inheritance from an older setup.py. I think we can remove it.

graingert: @andrewgodwin want me to land it in this PR?

andrewgodwin: Sure

andrewgodwin (approved):

PR #237, opened 2021-02-04
define SyncToAsync.thread_sensitive_context type

graingert:


graingert:
@andrewgodwin ^

andrewgodwin:
Thanks, this is certainly cleaner than _my_ attempt to fix this.

PR #239, opened 2021-02-09
Add type checks for functions passed to async_to_sync/sync_to_async

### LINKED ISSUE #238: Prevent accidental mixup of sync_to_async / async_to_sync
Putting `sync_to_async` onto an `async def` silently breaks it. Instead, it would be good if that raised an error at decorator application time - and likewise for the opposite case.

adamchainz:
Fixes #238.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,9 @@
+Pending
+-------
+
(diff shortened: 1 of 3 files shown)


andrewgodwin:
Lovely, thanks!

PR #242, opened 2021-02-26
simplify example implementation of the lifespan protocol

### LINKED ISSUE #241: [Docs] Lifespan ref implementation
Hi everyone. Looking at https://asgi.readthedocs.io/en/latest/specs/lifespan.html , I saw the following example implementation of the `lifespan` type.

```
        while True:
            message = await receive()
            if message['type'] == 'lifespan.startup':
                ... # Do some startup here!
                await send({'type': 'lifespan.startup.complete'})
            elif message['type'] == 'lifespan.shutdown':
                ... # Do some shutdown here!
                await send({'type': 'lifespan.shutdown.complete'})
                return
```

My naive implementation would consist of four flat lines of code only (receive, send, receive, send).

What is the reason for the `while True` loop and the fail-safe message handling here? It's seems like I am missing something.

srkunze:
Should fix #241 

I leave it here for discussion.

### DIFF
--- a/specs/lifespan.rst
+++ b/specs/lifespan.rst
@@ -18,15 +18,15 @@ A possible implementation of this protocol is given below::
 
     async def app(scope, receive, send):
         if scope['type'] == 'lifespan':
-            while True:
-                message = await receive()
-                if message['type'] == 'lifespan.startup':
-                    ... # Do some startup here!
-                    await send({'type': 'lifespan.startup.complete'})
-                elif message['type'] == 'lifespan.shutdown':
-                    ... # Do some shutdown here!
-                    await send({'type': 'lifespan.shutdown.complete'})
-                    return
+            message = await receive()
+            assert message['type'] == 'lifespan.startup'
+            ... # Do some startup here!
+            await send({'type': 'lifespan.startup.complete'})
+
+            message = await receive()
+            assert message['type'] == 'lifespan.shutdown'
+            ... # Do some shutdown here!
+            await send({'type': 'lifespan.shutdown.complete'})
         else:
             pass # Handle other types
 


sm-Fifteen:
As far as I'm aware, this example would be insufficient, since unlike every other ASGI protocol, lifespan handlers are [forbidden from throwing](https://asgi.readthedocs.io/en/latest/specs/lifespan.html#scope) in their normal execution flow:

 > If an exception is raised when calling the application callable with a lifespan.startup message or a scope with type lifespan, the server must continue but not send any lifespan events.
> 
> This allows for compatibility with applications that do not support the lifespan protocol. If you want to log an error that occurs during lifespan startup and prevent the server from starting, then send back lifespan.startup.failed instead.

If your application does support `lifespan`, you probably don't want the server to ignore that error and run the application like normal.

Also, ASGI apps works by waiting for the server to signal them (via `receive()`) that something happened (such as `"http.request"`: "an HTTP client wants something" or `'lifespan.shutdown'`: "The server is about to shutdown"), doing something in response and signalling the server (via `send()`) in return. Everything needs to be able to happen in somewhat of a disjointed manner, even lifespan event handling, hence the loop and the startup and shutdown being in their own branches instead of one after the other.

It should look more like this:

```py
    while True:
        message = await receive()

        if scope['type'] == 'lifespan':
            # All exceptions must be caught while handling a lifespan event if lifespan is supported.
            if message['type'] == 'lifespan.startup':
                try:
                    ... # Do some startup here!
                    await send({'type': 'lifespan.startup.complete'})
                except Exception as e:
                    await send({'type': 'lifespan.startup.failed', 'message': str(e)})

            elif message['type'] == 'lifespan.shutdown':
                try:
                    ... # Do some shutdown here!
                    await send({'type': 'lifespan.shutdown.complete'})
                except Exception as e:
                    await send({'type': 'lifespan.shutdown.failed', 'message': str(e)})
            else:
                # Invalid or unsupported type, applications should raise an exception to the server.
                raise NotImplementedError()

        elif scope['type'] == 'http':
            ... # Handle http
```

PR #244, opened 2021-03-04
Parameterize executor

joshuahaertel:
There are cases where it may be desirable to run certain sync_to_async callables in their own executor. For example, a user might want to allow a method to be run in small thread pool without affecting the loop's default thread pool. Or, a custom executor may be passed in to gather metrics about calls for certain decorated methods.

### DIFF
--- a/asgiref/sync.py
+++ b/asgiref/sync.py
@@ -5,7 +5,7 @@ import sys
 import threading
 import weakref
 from concurrent.futures import Future, ThreadPoolExecutor
-from typing import Dict, Optional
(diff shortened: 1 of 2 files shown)


joshuahaertel:
@andrewgodwin Thoughts on this proposal?

andrewgodwin:
It's in my backlog, I'll get to this next week probably.

andrewgodwin (approved):

andrewgodwin:
Looks lovely, thanks!

PR #257, opened 2021-04-16
Use tox-py in CI

adamchainz:
This is a [tox plugin I just made](https://pypi.org/project/tox-py/) to replace this shell-fu.


andrewgodwin:
Still surprised this wasn't in normal `tox`, but looks good to me! Just don't break tox-py now :)

adamchainz:
It will lobby for it in the rewrite.

I won't break tox-py, I promise.

Issue #254, opened 2021-04-07
Bad type hints for async_to_sync and sync_to_async


adamchainz:
Take this example test program:

```python
import asyncio

from asgiref.sync import sync_to_async


def foo(x: int) -> None:
    print(x)


foo_async = sync_to_async(foo)


async def main() -> None:
    await foo_async(1)


asyncio.run(main())
```

On asgiref 3.3.2+ mypy complains:

```
$ mypy test.py
test.py:14: error: Incompatible types in "await" (actual type "Union[Any, SyncToAsync]", expected type "Awaitable[Any]")
test.py:14: error: Argument 1 has incompatible type "int"; expected "Callable[..., Any]"
Found 2 errors in 1 file (checked 1 source file)
```

Similar errors appear with the reverse:

```python
from asgiref.sync import async_to_sync


async def foo(x: int) -> None:
    print(x)


foo_sync = async_to_sync(foo)


def main() -> None:
    foo_sync(1)


main()
```

Pinging @graingert who added the type annotations in #233 and is generally a mypy wizard.

FinnStutzenstein:
I have the same issue.

graingert:
"Union[Any, SyncToAsync]" should absolutely be compatible with expected type "Awaitable[Any]"

This is generally unfixable without ParamSpec support in mypy though

graingert:
type annotations were added here: https://github.com/django/asgiref/commit/b3257f6ca844ddb303f4d10f5dd66d86ecf43200#diff-6af8b5512d949418a62aa4e6be5c741b376f69feacf7cc5e3fe57162aaba1650R472

an `@overload` would probably help here

andrewgodwin:
What's the best fix here? I don't want to remove the annotations but unless we can come up with something a bit more compatible I feel like we'll have to at least temporarily to stop mypy blowing up.

adamchainz:
Should `SyncToAsync` inherit from `collections.abc.Awaitable` ?

graingert:
You could use an alphabet soup `@overload`


```
@overload
def async_to_sync(fn: Callable[[A,B,C,D,E,F,G,H,I,J,K,L,M,N,O,P,Q,R,S,T,U,V,W,X,Y,Z], Awaitable[ZZ]]) -> Callable[[A,B,C,D,E,F,G,H,I,J,K,L,M,N,O,P,Q,R,S,T,U,V,W,X,Y,Z], ZZ]): ...`
```

PR #261, opened 2021-05-20
Fix type hints for sync_to_async()

adamchainz:
Use `@overload` to specify the two different sets of types for `sync_to_async()`. This prevents any errors in the example shown in #254.

### DIFF
--- a/asgiref/sync.py
+++ b/asgiref/sync.py
@@ -7,7 +7,7 @@ import threading
 import warnings
 import weakref
 from concurrent.futures import Future, ThreadPoolExecutor
-from typing import Any, Callable, Dict, Optional, Union
+from typing import Any, Callable, Dict, Optional, overload
 
 from .current_thread_executor import CurrentThreadExecutor
 from .local import Local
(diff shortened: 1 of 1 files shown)


adamchainz:
@graingert any input?

andrewgodwin:
Thanks for this @adamchainz! I'll get a release out once I have time to thoroughly test #262.

pawelrubin:
This is neat! 🔥 

@andrewgodwin What are the odds of releasing this in the near future? 🙏🏻

andrewgodwin:
I've been waiting to get a bit more testing done on #262 with python 3.6 and 3.10 on both Flask and Django - if someone wants to run through that and make sure everything seems superficially happy for me I can get the release out very soon, otherwise I need to wait until I can sit down and double-check that.

Mng-dev-ai:
@adamchainz I think the first type of for sync_to_async() should be like this : 
```
@overload
def sync_to_async(
    func: None = None,
    thread_sensitive: bool = True,
    executor: Optional["ThreadPoolExecutor"] = None,
) -> Callable[[Callable[..., Any]], Awaitable]:

```
so the main difference here that we return Awaitable instead of SyncToAsync

PR #262, opened 2021-05-20
Fix deprecation warnings for python 3.10

vaartis:
asyncio.get_event_loop was marked as deprecated, the documnetation
now refers to asyncio.get_running_loop([1])

asyncio.ensure_future issues a deprecation warning if there is no
running event loop([2]), so use asyncio.run which creates and destroys the
loop itself

asyncio.gather issues a warning if run outside of event
loop (i.e. there is no running event loop)([3]), so wrap it into an
async def

explicit passing of coroutine objects to asyncio.wait is deprecated
since 3.8([4]), so wrap them in asyncio.create_task

plus, add 3.10 to tox.ini

[1]: https://docs.python.org/3.10/library/asyncio-eventloop.html#asyncio.get_event_loop
[2]: https://docs.python.org/3.10/library/asyncio-future.html#asyncio.ensure_future
[3]: https://docs.python.org/3.10/library/asyncio-task.html#asyncio.gather
[4]: https://docs.python.org/3.10/library/asyncio-task.html#asyncio.wait

### DIFF
--- a/asgiref/compatibility.py
+++ b/asgiref/compatibility.py
@@ -1,5 +1,6 @@
 import asyncio
 import inspect
+import sys
 
 
 def is_double_callable(application):
@@ -45,3 +46,16 @@ def guarantee_single_callable(application):
     if is_double_callable(application):
         application = double_to_single_callable(application)
     return application
+
+
+if sys.version_info >= (3, 7):
+    # these were introduced in 3.7
+    get_running_loop = asyncio.get_running_loop
+    run_future = asyncio.run
+    create_task = asyncio.create_task
+else:
+    # marked as deprecated in 3.10, did not exist before 3.7
+    get_running_loop = asyncio.get_event_loop
+    run_future = asyncio.ensure_future
+    # does nothing, this is fine for <3.7
+    create_task = lambda task: task
--- a/asgiref/server.py
+++ b/asgiref/server.py
@@ -3,7 +3,7 @@ import logging
 import time
 import traceback
 
-from .compatibility import guarantee_single_callable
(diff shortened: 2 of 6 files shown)


carltongibson (commented):
Hi @vaartis, `get_running_loop()` was not added until Python 3.7 so there needs to be a bigger discussion about Python support before we can move to it. I’ve raised it as a possibility on #260. Let’s see what folks say. Thanks.

andrewgodwin (changes requested):
`asgiref` is required to work on Python 3.6 due to Django and Flask.

andrewgodwin (approved):
Lovely, I much prefer this. I'll merge it once the CI passes.

andrewgodwin:
Thanks for this! Were you testing this against Django, Flask, or something else? Due to the nature of the change I'll have to manually run some tests to make sure we didn't introduce subtle breakage and it'd be nice to know if some of those areas are already covered.

vaartis:
> Thanks for this! Were you testing this against Django, Flask, or something else? Due to the nature of the change I'll have to manually run some tests to make sure we didn't introduce subtle breakage and it'd be nice to know if some of those areas are already covered.

Django, at least the test suite, did pass. However I didn't test anything on 3.6, only 3.8 and 3.10.

andrewgodwin:
> Django, at least the test suite, did pass. However I didn't test anything on 3.6, only 3.8 and 3.10.

Great, thanks for that. I'm not too worried about 3.6 since this patch barely changes behaviour for it - I'll run a few extra tests with some weird async Django apps I have to make sure and then I'll get this into a release.

PR #263, opened 2021-05-26
Raise on deadlock

shughes-uk:
Raise an exception on potential deadlock where the single thread pool is awaited on while already being awaited on in the same context 

I'm using a context var here because I think setting an attribute would throw an exception outside of this scenario, if you had multiple 'main' threads just wanting to use the single thread executor, which would not deadlock

### DIFF
--- a/asgiref/sync.py
+++ b/asgiref/sync.py
@@ -348,6 +348,15 @@ class SyncToAsync:
     else:
         thread_sensitive_context: None = None
 
+    # Contextvar that is used to detect if the single thread executor
+    # would be awaited on while already being used in the same context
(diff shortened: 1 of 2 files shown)


shughes-uk:
Added a `pytest.mark.skipif` for python 3.6 as this solution only works in 3.7 due to the use of contextvars. Is this acceptable or is a resolution compatible with 3.6 be implemented?

andrewgodwin:
It's fine to make it only error on 3.7 - this is extra protection, rather than a core feature.

I think the patch looks good! I'll merge it in, and thanks!

PR #271, opened 2021-06-25
🐛 Fix extensions type on HTTPScope and WebsocketScope

Kludex:
@andrewgodwin We are annotating `uvicorn` with `asgiref` now, and possibly extending this idea to `Starlette`. So... Thanks for maintaining it! :)

While I was annotating `uvicorn`, I've noticed that `mypy` complains about the nonexistence of "extensions" so I added it. But after looking at the ASGI specs, it doesn't look like "extensions" is mandatory. Am I wrong?

As per #221 it was decided to use `Optional` to optional fields, instead of making the field optional _per se_, so I've created this PR.


andrewgodwin:
Extensions is optional, so this seems correct. I'll land it provided the checks pass!

andrewgodwin:
Great, thanks for this!

PR #272, opened 2021-06-25
📝 Rename Websocket by WebSocket

Kludex:
This is fundamentally unnecessary, so the rejecting this PR is understandable. But the correct name is WebSocket.

### DIFF
--- a/asgiref/_pep562.py
+++ b/asgiref/_pep562.py
@@ -42,7 +42,9 @@ class Pep562:
         return self._get_dir() if self._get_dir else dir(self._module)
 
     def __getattr__(self, name: str) -> Any:
-        """Attempt to retrieve the attribute from the module, and if missing, use the overridden function if present."""
+        """
+        Attempt to retrieve the attribute from the module, and if missing, use the overridden function if present.
+        """
 
         try:
             return getattr(self._module, name)
--- a/asgiref/_pep562.py
+++ b/asgiref/_pep562.py
@@ -1,0 +1,59 @@
+"""
+Backport of PEP 562.
+https://pypi.org/search/?q=pep562
+Licensed under MIT
+Copyright (c) 2018 Isaac Muse <isaacmuse@gmail.com>
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+documentation files (the "Software"), to deal in the Software without restriction, including without limitation
+the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software,
+and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+The above copyright notice and this permission notice shall be included in all copies or substantial portions
+of the Software.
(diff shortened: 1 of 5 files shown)


andrewgodwin:
I don't object to it, but if you want to land this you'll need to do it in a backwards-compatible fashion (so the old names work but raise a warning)

Kludex:
I've used [PEP 562](https://www.python.org/dev/peps/pep-0562/) idea here, but as it was introduced in 3.7, I've used the [backport implementation for 3.6](https://github.com/facelessuser/pep562).

I've also added tests to check that the warning is being raised on the wanted types.

andrewgodwin:
I'd rather just do it the "old fashioned" way if we could - expose them as actual top-level module objects rather than bringing in extra code just for this.

Kludex:
I'm not following you. :thinking: How am I supposed to warn the user about the deprecation at import time without this implementation?

Or you just meant to not use `__all__` and `__deprecated__`?

andrewgodwin:
Oh, I forgot they were type classes for a second, sorry. Thought they were actual classes where you could do it with a mixin on instantiation.

In that case, let's vendor pep562 in provided the license matches - I don't want more dependencies!

asgiref/_pep562.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def __init__(self, name: str) -> None:
+        """Acquire `__getattr__` and `__dir__`, but only replace module for versions less than Python 3.7."""
+
+        self._module = sys.modules[name]
+        self._get_attr = getattr(self._module, "__getattr__", None)
+        self._get_dir: Optional[Callable[..., List[str]]] = getattr(
+            self._module, "__dir__", None
+        )
+        sys.modules[name] = self  # type: ignore[assignment]
```
Kludex, line 37: `sys.modules` expects a `Module` and `self` is a `Pep562`, jfyk.

Kludex:
There's a mismatch between the linters, `isort` and `black` are reformatting themselves on the `pre-commit` :thinking:

Kludex:
Notes about the changes: 
- I've added `pep562` as private module, so we can remove easily by the [3.6 EOL](https://endoflife.date/python).
- I've removed all the unnecessary code on `_pep562.py` (and tried to modify the less I could, jfyk).

andrewgodwin:
I think sticking with `black`'s default length of 88 is the best approach - let's handle that in a separate PR. I'll merge this in now so we can avoid mixing linting in here - I think it looks good now, so thanks for bearing with my requests!

Kludex:
Thank you! :tada:

Kludex:
This is not `mypy` compliant. A project with `asgiref` not pinned, will receive this on the upgrade:
```python
+ mypy --show-error-codes
uvicorn/middleware/proxy_headers.py:13: error: Module "asgiref.typing" has no attribute "WebsocketScope"; maybe "WebSocketScope"?  [attr-defined]
```

Which I think it's expected. but I'm mentioning because I'm not sure you were aware. @andrewgodwin

PR #274, opened 2021-06-27
:hammer: Match isort line length to black

Kludex:


Kludex:
You are too fast! :nerd_face:

andrewgodwin:
Apparently also not correct enough, as I had to sneak https://github.com/django/asgiref/commit/cc7b61881f2c1ea0d01bc77d83afa96aa15906d8 in as well.

PR #282, opened 2021-07-15
New sync_to_async tests, one of which triggers a deadlock

asedeno:
These tests provide an example the issues I ran into while moving from asgiref 3.2.10 to 3.3+, which digging through issues here has suggested was not uncommon. Making `sync_to_async` default to `thread_sensitive=True` can cause deadlocks in previously functioning code. I've read and understood why the choice was made, and don't presently have any suggestions on how to better detect and avoid the situation, but figured the distilled scenario in test form might be useful regardless.

---
In these tests, we are waiting on a `mulitprocessing.Event` to be set, with the wait in `sync_to_async()`.

The event is being set in an `asyncio.Task` that sleeps and then calls set, also in `sync_to_async()`. This allows us to kick off the process that would `set()` the event in advance of `wait()`ing for the event, as the event might be triggered by some consumer event.

If both `sync_to_async()` calls are `thread_sensitive=True`, then before `await sync_to_async(event.set)()` can run `event.set()`, `sync_to_async(event.wait)()` has consumed the thread.

asgiref.timeout is used to get us out of the deadlock so the tests can run to completion, and in the case of Timeout we `set()` the event to release the `wait()`.

Finally, we await on the setting task regardless of whether or not there was a timeout.

### DIFF
--- a/tests/test_sync.py
+++ b/tests/test_sync.py
@@ -728,7 +728,7 @@ async def test_sync_to_async_with_blocker_thread_sensitive():
     Expected to fail at the moment.
     """
 
-    delay = 1 # second
+    delay = 1  # second
     event = multiprocessing.Event()
 
     async def async_process_waiting_on_event():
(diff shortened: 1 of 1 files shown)


andrewgodwin:
This looks great - a `xfail` test is the perfect place to start from to fix the deadlock detection/issues. Get 3.6 happy and the linting passing and I'll happily take this (and hopefully it'll help fix the problems)

asedeno:
py3.6 and black are good now. 👍

andrewgodwin:
Great, thanks so much!

PR #283, opened 2021-07-15
fix `async_to_async` typo

asedeno:
Something I noticed while working on #282, but which is not part of it.


andrewgodwin:
Nice catch!

PR #291, opened 2021-08-16
timeout module: Hoist deciding which asyncio current_task to use to module definition time

kezabelle:
Follows up on f53fbef (#289). Note that the module level variable used here differs, because it is already in use as the public function name.

As with that one, saves 1 `hasattr` call per usage, this time of of `asgiref.timeout.current_task`

### DIFF
--- a/asgiref/timeout.py
+++ b/asgiref/timeout.py
@@ -7,9 +7,10 @@
 
 
 import asyncio
-import sys
 from types import TracebackType
 from typing import Any, Optional, Type
+
+from .compatibility import current_task as asyncio_current_task
 
 
 class timeout:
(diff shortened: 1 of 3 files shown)


asgiref/timeout.py
```diff
@@ -12,6 +12,12 @@
 from typing import Any, Optional, Type
 
 
+if sys.version_info >= (3, 7):
+    _current_task = asyncio.current_task
```
andrewgodwin, line 16: Could we use a hasattr check here? Just something I prefer.

andrewgodwin:
Looks good, just one comment (and obviously the linting fix)

kezabelle:
So before I make the change (happy to do so), I'll throw out another option in case you're amenable to that.

 Upon further poking around of the package, I see there's a `compatibility` module with the same `sys.version_info >= (3, 7)` situation, which is used by the `sync` and `server` modules. I could just bung the current_task detection in there (leaving it using the version_info check _or_ having it be a separate `hasattr` as you indicated), and then import it into `timeout` and `sync` as necessary. I'd still have to alias it to `_current_task` or suchlike for the `timeout` module (this PR's target), and it would retroactively obviate the previous PR a bit, but it'd consolidate things to one location rather than 2 which is maybe a boon given the lifetime of 3.6 support is probably quite long?

andrewgodwin:
Oh yes, this was in fact the point of that part of `compatibility`, so I think shoving this and the previous PR's target in there is probably the right call!

PR #293, opened 2021-08-18
Removed Local.CLEANUP_INTERVAL class variable which has been unused for a while

kezabelle:
It was last used prior to a0bbe901ffe19cd927219989b7125a920ea1e5c3 and tags < `3.2.6`
It may be going through an informal deprecation cycle, or it may've just been missed (wood for the trees etc). Either way, when it's appropriate, this should be OK to merge. One hopes.

### DIFF
--- a/asgiref/local.py
+++ b/asgiref/local.py
@@ -30,8 +30,6 @@ class Local:
     3.7 only, we can then reimplement the storage more nicely.
     """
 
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Yup, must have missed this when we moved from time-based eviction! Thanks!

PR #294, opened 2021-08-30
use heapq to remove old application instances efficiently

bright2227:
previous function removing one instance at a time.  One for loop to find oldest last_used time,  followed by another loop to find the instance with oldest last_used time.  These steps should be combined together.   
Heapq is good at case to find nth highest or lowest data.

### DIFF
--- a/asgiref/server.py
+++ b/asgiref/server.py
@@ -2,6 +2,7 @@ import asyncio
 import logging
 import time
 import traceback
+from heapq import heappush, heappushpop
 
 from .compatibility import get_running_loop, guarantee_single_callable, run_future
 
@@ -82,8 +83,9 @@ class StatelessServer:
             self.application_instances[scope_id]["last_used"] = time.time()
             return self.application_instances[scope_id]["input_queue"]
         # See if we need to delete an old one
-        while len(self.application_instances) > self.max_applications:
-            self.delete_oldest_application_instance()
+        remove_num = len(self.application_instances) - self.max_applications
+        if remove_num > 0:
+            self.delete_nth_old_application_instance(remove_num)
         # Make an instance of the application
         input_queue = asyncio.Queue()
         application_instance = guarantee_single_callable(self.application)
@@ -103,19 +105,19 @@ class StatelessServer:
         }
         return input_queue
 
-    def delete_oldest_application_instance(self):
+    def delete_nth_old_application_instance(self, remove_num):
         """
-        Finds and deletes the oldest application instance
+        Finds and deletes the nth old application instance with heapq
         """
-        oldest_time = min(
-            details["last_used"] for details in self.application_instances.values()
-        )
+        nth_old_instances = []
         for scope_id, details in self.application_instances.items():
-            if details["last_used"] == oldest_time:
-                self.delete_application_instance(scope_id)
-                # Return to make sure we only delete one in case two have
-                # the same oldest time
-                return
+            if len(nth_old_instances) < remove_num:
+                heappush(nth_old_instances, (details["last_used"], scope_id))
+            else:
+                heappushpop(nth_old_instances, (details["last_used"], scope_id))
+
+        for _, scope_id in nth_old_instances:
+            self.delete_application_instance(scope_id)
 
     def delete_application_instance(self, scope_id):
         """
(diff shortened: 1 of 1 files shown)


asgiref/server.py
```diff
@@ -134,7 +136,7 @@ async def application_checker(self):
         """
         while True:
             await asyncio.sleep(self.application_checker_interval)
-            for scope_id, details in list(self.application_instances.items()):
+            for scope_id, details in self.application_instances.items():
```
andrewgodwin, line 139: Have you checked to make sure this still works under mutation? That's why the `list()` was there in the first place.

andrewgodwin (commented):
I like the idea, but I'd want to see two more things:
* A comparison benchmark before and after the change - it's adding complexity, so I want to make sure it's enough of an improvement considering the generally small size of the datastructure
* A test or two that covers this area, since we don't already have one and I want to make sure this doesn't beak anything

asgiref/server.py
bright2227: The new change passed the testing of pytest.  I made this change just because the similar method is called without list() constructor in the previous delete_oldest_application_instance method.

```
    def delete_oldest_application_instance(self):
        """
        Finds and deletes the oldest application instance
        """
        oldest_time = min(
            details["last_used"] for details in self.application_instances.values()
        )
        for scope_id, details in self.application_instances.items():
            if details["last_used"] == oldest_time:
                self.delete_application_instance(scope_id)
                # Return to make sure we only delete one in case two have
                # the same oldest time
                return
```

bright2227:
I think you're right about building testing.  I didn't notice that the original test is too simple. If I use the original code and add one more line in testing,  it rises error.

```
from asgiref.server import StatelessServer


def test_stateless_server():
    """StatelessServer can be instantiated with an ASGI 3 application."""

        pass

    server = StatelessServer(app)
    server.get_or_create_application_instance("scope_id", {})
    server.get_or_create_application_instance("scope_id2", {})   # new line
```
It causes errors below.
```
__________________________________________________________ test_stateless_server __________________________________________________________

    def test_stateless_server():
        """StatelessServer can be instantiated with an ASGI 3 application."""
    
            pass
    
        server = StatelessServer(app)
        server.get_or_create_application_instance("scope_id", {})
>       server.get_or_create_application_instance("scope_id2", {})

tests/test_server.py:12: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
asgiref/server.py:82: in get_or_create_application_instance
    input_queue = asyncio.Queue()
/usr/lib/python3.8/asyncio/queues.py:35: in __init__
    self._loop = events.get_event_loop()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <asyncio.unix_events._UnixDefaultEventLoopPolicy object at 0x7f66bc73de80>

    def get_event_loop(self):
        """Get the event loop for the current context.
    
        Returns an instance of EventLoop or raises an exception.
        """
        if (self._local._loop is None and
                not self._local._set_called and
                isinstance(threading.current_thread(), threading._MainThread)):
            self.set_event_loop(self.new_event_loop())
    
        if self._local._loop is None:
>           raise RuntimeError('There is no current event loop in thread %r.'
                               % threading.current_thread().name)
E           RuntimeError: There is no current event loop in thread 'MainThread'.

/usr/lib/python3.8/asyncio/events.py:639: RuntimeError
```
I'm going to build a proper test first and perhaps remove the potential bugs.

asgiref/server.py
bright2227: Oh, sorry. I find that list() constructor is meant to avoid dictionary changed size during iteration.

bright2227:
Excuse me. I'm not quite familiar with git.  How I made changes based on recent commits? Should I restart a new pull request?

PR #298, opened 2021-09-25
Typing for sync/async decorators

shughes-uk:
This is a first pass draft, fairly sure it's broken for python versions lower than 3.8. 

Seems to work but i'm not sure how to test it right this moment, i'm used to using pyright and will have to investigate how to use mypy

One thing that confused me a bit is that the async_to_sync function accepts `awaitable` but it actually seems to be looking for a callable that returns an awaitable? I could use some clarification there.

I'm also not sure what naming convention to use for the TypeVar/Paramspec pairs.

Looks like I also introduce some formatting errors to correct!

### DIFF
--- a/asgiref/sync.py
+++ b/asgiref/sync.py
@@ -7,7 +7,7 @@ import threading
 import warnings
 import weakref
 from concurrent.futures import Future, ThreadPoolExecutor
-from typing import Any, Awaitable, Callable, Dict, Generic, Optional, TypeVar, overload
+from typing import Any, Awaitable, Callable, Dict, Optional, TypeVar, Union, overload
 
 from typing_extensions import ParamSpec
 
@@ -19,6 +19,9 @@ if sys.version_info >= (3, 7):
     import contextvars
 else:
     contextvars = None
+
+P = ParamSpec("P")
+R = TypeVar("R")
 
 
 def _restore_context(context):
@@ -100,11 +103,7 @@ class ThreadSensitiveContext:
             pass
 
 
-a_cls_params = ParamSpec("a_cls_params")
-a_cls_return = TypeVar("a_cls_return")
-
-
-class AsyncToSync(Generic[a_cls_params, a_cls_return]):
+class AsyncToSync:
     """
     Utility class which turns an awaitable that only works on the thread with
     the event loop into a synchronous callable that works in a subthread.
@@ -125,9 +124,7 @@ class AsyncToSync(Generic[a_cls_params, a_cls_return]):
     executors = Local()
 
     def __init__(
-        self,
-        awaitable: Callable[a_cls_params, Awaitable[a_cls_return]],
-        force_new_loop=False,
+        self, awaitable: Callable[..., Awaitable[Any]], force_new_loop: bool = False
     ):
         if not callable(awaitable) or not _iscoroutinefunction_or_partial(awaitable):
             # Python does not have very reliable detection of async functions
@@ -137,7 +134,7 @@ class AsyncToSync(Generic[a_cls_params, a_cls_return]):
             )
         self.awaitable = awaitable
         try:
-            self.__self__ = self.awaitable.__self__
+            self.__self__ = self.awaitable.__self__  # type: ignore
         except AttributeError:
             pass
         if force_new_loop:
@@ -161,9 +158,7 @@ class AsyncToSync(Generic[a_cls_params, a_cls_return]):
                 else:
                     self.main_event_loop = None
 
-    def __call__(
-        self, *args: a_cls_params.args, **kwargs: a_cls_params.kwargs
-    ) -> a_cls_return:
+    def __call__(self, *args, **kwargs):
         # You can't call AsyncToSync from a thread with a running event loop
         try:
             event_loop = get_running_loop()
@@ -184,7 +179,7 @@ class AsyncToSync(Generic[a_cls_params, a_cls_return]):
             context = None
 
         # Make a future for the return information
-        call_result: Future[a_cls_return] = Future()
+        call_result = Future()
         # Get the source thread
         source_thread = threading.current_thread()
         # Make a CurrentThreadExecutor we'll use to idle in this thread - we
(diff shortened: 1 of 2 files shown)


shughes-uk:
Seems I have some work to do with mypy and this!

andrewgodwin:
No worries - take the time you need to get everything working! I might suggest you mark the PR as a draft, though, so that way you can move it to "ready for review" when you think you've got it all ready to go.

shughes-uk:
mypy seems not to support paramspec yet so this might take a little while unless you want to switch to pyright or pyre https://github.com/python/mypy/issues/8645

andrewgodwin:
I'm not really feeling like switching typechecker right at this moment, unfortunately. Sorry for delaying this yet further!

asgiref/sync.py
```diff
@@ -269,7 +281,13 @@ def __get__(self, parent, objtype):
         return functools.update_wrapper(func, self.awaitable)
 
     async def main_wrap(
-        self, args, kwargs, call_result, source_thread, exc_info, context
+        self,
+        args,
+        kwargs,
```
graingert, line 286: you need to type annotate all the parameters for mypy to check this function

It's probably worth making all the parameters positional only so you can use `*args` and `**kwargs`:


```python
def _main_wrap(call_result: Future[R], source_thread: threading.Thread, exc_info: ExcInfo, context: Optional[Context]) -> Callable[P, Coroutine[Any, Any, None]:
    async def wrap(*args: P.args, **kwargs: P.kwargs) -> None:
        ...

    return wrap
```

shughes-uk:
Not entirely sure whats going on but the checks appears to have been running for many hours 😱

graingert:
> Future is only generic in 3.9+ and will throw exceptions in previous versions, removed it for now

You'll need to use a forward reference with `"Future[R]"`

asgiref/sync.py
```diff
@@ -319,6 +318,10 @@ async def main_wrap(
                 _restore_context(context[0])
 
             current_task = SyncToAsync.get_current_task()
+            if not current_task:
```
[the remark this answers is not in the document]
shughes-uk: ah perfect thank you

shughes-uk:
Tests are passing and these are the mypy issues left. Most of which seem to be caused by mypy not recognizing Generic's can accept ParamSpecs

```
asgiref/sync.py:126: error: Free type variable expected in Generic[...]
asgiref/sync.py:159: error: "Callable[..., Awaitable[R]]" has no attribute "__self__"
asgiref/sync.py:183: error: Name "P.args" is not defined
asgiref/sync.py:183: error: Name "P.kwargs" is not defined
asgiref/sync.py:310: error: Name "P.args" is not defined
asgiref/sync.py:311: error: Name "P.kwargs" is not defined
asgiref/sync.py:348: error: Free type variable expected in Generic[...]
asgiref/sync.py:432: error: Name "P.args" is not defined
asgiref/sync.py:432: error: Name "P.kwargs" is not defined
asgiref/sync.py:441: error: Argument 1 to "get" of "ContextVar" has incompatible type "None"; expected "Union[ContextVar[str], str]"
asgiref/sync.py:498: error: Returning Any from function declared to return "R"
asgiref/sync.py:506: error: Function is missing a return type annotation
asgiref/sync.py:512: error: Name "P.args" is not defined
asgiref/sync.py:513: error: Name "P.kwargs" is not defined
asgiref/sync.py:569: error: "SyncToAsync" expects 1 type argument, but 2 given
asgiref/sync.py:569: error: Invalid location for ParamSpec "P"
asgiref/sync.py:569: note: You can use ParamSpec as the first argument to Callable, e.g., 'Callable[P, int]'
asgiref/sync.py:578: error: "SyncToAsync" expects 1 type argument, but 2 given
asgiref/sync.py:578: error: Invalid location for ParamSpec "P"
asgiref/sync.py:578: note: You can use ParamSpec as the first argument to Callable, e.g., 'Callable[P, int]'
```

shughes-uk:
@andrewgodwin I've updated this with a workaround based on the current (unreleased) mypy version. Temporarily installing mypy from the git version. Once mypy cuts a release this should be good to go. No generics required, but maybe a little hacky. 

Other option is to wait for generic support in mypy and a release. Which who knows maybe that'll happen before the next release rendering the workaround redundant.

andrewgodwin:
Will current versions of mypy-in-the-wild choke on the new stuff, btw, or will they ignore it and not typecheck it? Considering the type annotations here are mostly for the benefit of users rather than asgiref itself, I want to ensure it's not going to torpedo current users.

graingert:
Sadly mypy 0.910 will choke on this. I think the work-around is to provide these aliases in a new module eg `asgiref.sync_paramspec`

shughes-uk:
Yes this will blow up on wild mypy. Once they roll out a new release this could be merged without too much worry unless you intend for backwards compatibility with older mypy versions.

andrewgodwin:
Ugh, yeah, this is tricky because end-users mypy installations is the thing we care about at the end of the day. No way to somehow erase ParamSpec and fall back to Any or something on those older versions, I'm guessing?

setup.cfg
```diff
@@ -38,7 +38,7 @@ zip_safe = false
 tests =
     pytest
     pytest-asyncio
-    mypy>=0.800
+    mypy @ git+https://github.com/python/mypy.git@master
```
graingert, line 41:
```suggestion
    mypy >= 0.920
```

graingert: 0.920 was released yesterday http://mypy-lang.blogspot.com/2021/12/mypy-0920-released.html and supports ignoring ParamSpec in Callable[P, R] position

graingert:
mypy 0.930 is out with experimental support for ParamSpec!

lexicalunit:
Given that Python 3.6 is no longer supported as of asgiref version 3.5.0, is now a good time to revisit this discussion? Sorry, I've just been following along here and am really excited to see this feature land at some point 😄

andrewgodwin:
Maybe? I don't keep tight track of typing stuff right now so I'm not quite sure where things sit in terms of this working with 3.7.

graingert:
All versions of python support ParamSpec, even Python 2.7 all you need to do is import it in an "if TYPE_CHECKING:" block, or use the typing_extensions backport

graingert:
This got resolved in https://github.com/django/asgiref/pull/390

PR #308, opened 2021-12-27
StatelessServer now initializes the local event loop if necessary

alyjak:
fixes https://github.com/django/channels/issues/1713

### DIFF
--- a/asgiref/server.py
+++ b/asgiref/server.py
@@ -56,7 +56,8 @@ class StatelessServer:
         """
         Runs the asyncio event loop with our handler loop.
         """
-        event_loop = asyncio.get_event_loop()
+        policy = asyncio.get_event_loop_policy()
+        event_loop = policy.get_event_loop()
         asyncio.ensure_future(self.application_checker())
         try:
             event_loop.run_until_complete(self.handle())


andrewgodwin:
This is commented as being the same thing in the Python docs: https://docs.python.org/3.6/library/asyncio-eventloops.html#asyncio.get_event_loop

To avoid future confusion, could you add a test that fails with the previous code and works with this?

alyjak:
Hmm, this is confusing. Looking at the code for get_event_loop in [3.6](https://github.com/python/cpython/blob/3.6/Lib/asyncio/events.py#L682), [3.7](https://github.com/python/cpython/blob/3.7/Lib/asyncio/events.py#L739), and [3.10](https://github.com/python/cpython/blob/3.8/Lib/asyncio/events.py#L735) they all look identical minus the comment `# NOTE: this function is implemented in C (see _asynciomodule.c)` introduced in 3.7. Could it be that the c implementation of `_get_running_loop` introduced in 3.7 is raising a RuntimeError? The `.py` implementation of [_get_running_loop](https://github.com/python/cpython/blob/3.7/Lib/asyncio/events.py#L693) in 3.7 looks harmless.

Anyways, using the [minimal](https://github.com/bj00rn/asgiref-bug/tree/main) documented in [channels/issues/1713](https://github.com/django/channels/issues/1713), I'm able to demonstrate running a worker successfully when I change the asgiref requirement from 3.4.1 to my branch. This is using python3.9.5

alyjak:
K, I found the c impl and it looks like this might be the culprit: [`_asyncio_get_running_loop_impl(PyObject *module)`](https://github.com/python/cpython/blob/main/Modules/_asynciomodule.c#L3120). It appears this impl was introduced in 3.7. I've never read c python source, so I'm not sure what the difference is between that and [`_asyncio__get_running_loop_impl(PyObject *module)`](https://github.com/python/cpython/blob/3.7/Modules/_asynciomodule.c#L2980) (notice the double underscore) which appears to be the impl we would expect.

andrewgodwin:
Hm, yes, I wouldn't put it past CPython to change the underlying behaviour, async was still pretty new back then.

I'll take this in if there's a test for it - it's the sort of small change that might get refactored away again at some point otherwise.

alyjak:
Thanks for prodding me on the test. I found out that 4364f9b4bb052d9ec54b806867fd859b07c39186 actually fixed the issue and my tweak is essentially the same code. Prior to that commit `get_running_loop` was mapped by the compatibility module to `asyncio.get_event_loop` for py < 3.7 and mapped to `asyncio.get_running_loop` for py >= 3.7. [`asyncio.get_running_loop()`](https://github.com/python/cpython/blob/3.7/Lib/asyncio/events.py#L681) does raise a `RuntimeError` if no running loop is found.

It turns out that `tests/test_server.test_stateless_server` tests this issue. That test will fail with a `RuntimeError` if get_running_loop gets re-introduced to `StatelessServer.run()`. I've ensured the test passes with current code and fails if `asyncio.get_running_loop()` is used.

Note, if the thread-local loop is explicitly set to `None`, `StatelessServer.run()` will still fail with a RuntimeError with [`BaseDefaultEventLoopPolicy`](https://github.com/python/cpython/blob/3.7/Lib/asyncio/events.py#L632). I think this is expected behavior though, because I believe it means application code was responsible for removing the event loop. If it's not, you could catch the `RunTimeError` and explicitly set an new event loop using `policy.set_event_loop(policy.new_event_loop())`, which is what the BaseDefaultEventLoopPolicy actually does if its running in the main thread and the loop has never been set before.

andrewgodwin:
Oh, perfect, I just need to cut a release with all the recent changes then. Thanks for looking into this!

PR #312, opened 2022-01-24
Fix allowed values for `spec_version` on WebSocket and HTTP Connection scopes

Kludex:
Those values were not up-to-date considering what is on the top of the page:

image


andrewgodwin (approved):

andrewgodwin:
Lovely - thanks for catching that!

PR #319, opened 2022-03-11
Fix a Typographical Error.

ryomahan:
Fix Version History and Copyright level under Lifespan Protocol chapter.

### DIFF
--- a/specs/lifespan.rst
+++ b/specs/lifespan.rst
@@ -122,7 +122,7 @@ Keys:
 
 
 Version History
-''''''''''''''''''''''''''''''''
+'''''''''''''''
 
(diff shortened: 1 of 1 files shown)


specs/lifespan.rst
```diff
@@ -122,14 +122,14 @@ Keys:
 
 
 Version History
-===============
+''''''''''''''''''''''''''''''''
```
andrewgodwin, line 125: You need to make the length of these match the titles above them :) Once that's done, I'll happily merge it though - good catch.

ryomahan: Okay, I finished the revision.

andrewgodwin:
Great, thanks!

PR #320, opened 2022-03-18
Preserve CurrentThreadExecutor across create_task

### LINKED ISSUE #214: sync_to_async does not find root thread inside a task
When you use `create_task` (or any way of making a coroutine not through asgiref), and the root thread is synchronous, then `sync_to_async` in thread-sensitive mode will use the root thread outside of tasks and a single, specific new thread inside of tasks, ruining the thread-sensitive guarantee.

Instead, we should make `sync_to_async` always first look for an executor on the root thread and, if it finds it, use that to run code.

andersk:
Fixes #214.

### DIFF
--- a/asgiref/sync.py
+++ b/asgiref/sync.py
@@ -1,3 +1,4 @@
+import asyncio
 import asyncio.coroutines
 import contextvars
 import functools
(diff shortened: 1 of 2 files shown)


andersk:
@andrewgodwin Can I ask whether this approach makes sense for upstream? I understand if the actual merging and release might be slow for various reasons—but I want to make sure I won’t be stuck maintaining a permanent fork of asgiref for Zulip.

andrewgodwin:
Ah, sorry, this PR totally slipped out of my attention. It looks good!

andrewgodwin:
@andersk FYI, this is now out in 3.5.1 so you can hopefully undo any forking.

PR #321, opened 2022-03-31
Remove `SOCK_NONBLOCK` from socket creation on tests

ereslibre:
This fixes running tests on Mac OS X, which lacks the `SOCK_NONBLOCK`
socket flag creation.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
         args: ["--py37-plus"]
 
   - repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: 22.3.0
     hooks:
       - id: black
         args: ["--target-version=py37"]
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,7 +36,6 @@ zip_safe = false
 
 [options.extras_require]
 tests =
-    click==8.0.4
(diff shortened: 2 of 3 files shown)


andrewgodwin:
I wish I could remember _why_ it was there in the first place, but I can't think of a good reason and the tests seem to pass, so... sure?

You'll need to pin the version of Click to fix that linting error for now though - see https://github.com/psf/black/issues/2964

ereslibre:
@andrewgodwin I'm sorry I don't know what's the most appropriate way to pin click, not familiar with python/django development, and I don't see a conclusive way to do so out of the box. A pointer will do, thanks!

Kludex:
Here: https://github.com/django/asgiref/blob/cde961b13c69b90216c4c1c81d1ad1ca1bc22b48/setup.cfg#L37-L41

Add `click==<version_that_works>`

ereslibre:
Thank you @Kludex. Done.

andrewgodwin:
Looks like I got it wrong there - it was the pre-commit hook that needed fixing! I've made the changes for you - thanks for the PR!

ereslibre:
That was my first try, but I did undo it just because I didn't know if it was what you wanted :)

Thanks for the reviews!

PR #329, opened 2022-05-12
Fix `root_path` type in WebSocket Connection Scope

abersheeran:
This should be a Unicode String instead of a byte String as in HTTP.


Kludex:
Probably doesn't matter, as it's clearly a typo. But... Both Hypercorn and Uvicorn considered `root_path` as Unicode string already.

Kludex (approved):

PR #335, opened 2022-07-01
Use update_wrapper in SyncToAsync.__get__

legau:
Adds an `update_wrapper` in `SyncToAsync.__get__` to get more context out of the returned partial.
Same behavior as in `AsyncToSync`


andrewgodwin:
Thanks!

PR #338, opened 2022-07-23
Main

adriangb:
This tweaks the wording on the relationship between lifespans, event loops, threads and processes.

I believe this is not introducing anything new, the intention is already there in the spec, it's just not very strongly worded/clear.

All of the server implementations that are currently in widespread use and support lifespans (Hypercorn and Uvicorn; Daphne does not support lifespans) comply with this, so this will not break/conflict with any implementations.


andrewgodwin (approved):

adriangb:
Thank you, Andrew!

PR #339, opened 2022-08-14
Add HTTP Trailers extension

### LINKED ISSUE #141: Protocol: Support for HTTP trailers
It doesn't seem like the current ASGI specification allows applications to send HTTP trailers in their response, since it is indicated that [servers should ignore any further messages on a channel once `mode_body` becomes false](https://asgi.readthedocs.io/en/latest/specs/www.html#response-body) in a `http.response.body` message. This prevents application that stream their content from sending additionnal response information as trailers when it may not be known before the body has begun transfering, regarding things like caching, timing (`Server-Timing`), integrity check (`Content-MD5`, obsolete), digital signature (`Signature`, draft), etc.

WSGI doesn't support it and I'm reading that overall support for them in various user agents tends to be flaky most of the time, but I couldn't find any information regarding the omission of trailers from ASGI (whether deliberate or not) and I would like to know if there are any plans to add them at some point in the future, perhaps as an extension.

Kludex:
- Closes #141

Currently, there are two implementations of HTTP Trailer:
- https://github.com/encode/uvicorn/pull/1599/ (Uvicorn)
- https://github.com/nonebot/nonecorn/blob/nonecorn_dev/src/hypercorn/protocol/http_stream.py#L232 (Nonecorn - Hypercorn fork) - Commit that introduced it: https://github.com/nonebot/nonecorn/commit/222757f8da25fc5b716b790b93c23710942790b4

What this PR proposes doesn't match any of those two implementations. I've tried to simplify a bit...

The uvicorn one introduces a new key on the `http.response.start` called `trailers`, to show the intention from the application to the server that trailing headers will be sent. I think this is not needed. The server can understand the intention if the header `Trailer:` is provided on the `headers` from the `http.response.start`.

The nonecorn uses a different name for the extension, but it was the first one implemented. I just thought the name `http.response.trailers` was a bit better, but happy to change if wanted. cc @synodriver

~I understood less about the nonecorn implementation in comparison to the one from uvicorn, so I may be missing something here that I couldn't grasp from there. So I'd like your opinion here @synodriver. 🙏~

The `nonecorn` implementation makes the ASGI framework in charge of sending the `Trailer:` header, and it sets the `more_body` field to `True` on the last `http.response.body` message, and then it's able to use trailing headers message (called `http.trailingheaders.send` there) as last message, and then close the transport after it.  

cc @simonw @abersheeran @pgjones @synodriver @sam-kleiner @tomchristie 

References used to write this:
- https://fullstack.wiki/http/headers/Trailer
- https://fullstack.wiki/http/headers/TE
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Trailer
- https://chromium.googlesource.com/external/github.com/grpc/grpc/+/HEAD/doc/PROTOCOL-HTTP2.md

### DIFF
--- a/asgiref/typing.py
+++ b/asgiref/typing.py
@@ -16,6 +16,7 @@ __all__ = (
     "HTTPRequestEvent",
     "HTTPResponseStartEvent",
     "HTTPResponseBodyEvent",
+    "HTTPResponseTrailersEvent",
     "HTTPServerPushEvent",
     "HTTPDisconnectEvent",
     "WebSocketConnectEvent",
@@ -99,12 +100,19 @@ class HTTPResponseStartEvent(TypedDict):
     type: Literal["http.response.start"]
     status: int
     headers: Iterable[Tuple[bytes, bytes]]
+    trailers: bool
 
 
 class HTTPResponseBodyEvent(TypedDict):
     type: Literal["http.response.body"]
     body: bytes
     more_body: bool
+
+
+class HTTPResponseTrailersEvent(TypedDict):
+    type: Literal["http.response.trailers"]
+    headers: Iterable[Tuple[bytes, bytes]]
+    more_trailers: bool
 
 
 class HTTPServerPushEvent(TypedDict):
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -152,10 +152,14 @@ ASGI servers that implement this extension will provide
         },
     }
 
-An ASGI framework interested in sending trailing headers to the client, must include
-the ``Trailer`` header in the ``http.response.start``. That will allow the ASGI server
+An ASGI framework interested in sending trailing headers to the client, must set the
+field ``trailers`` in *Response Start* as ``True``. That will allow the ASGI server
 to know that after the last ``http.response.body`` message (``more_body`` being ``False``),
 the ASGI framework will send a ``http.response.trailers`` message.
+
+The ASGI framework is in charge of sending the ``Trailers`` headers to let the client know
+which trailing headers the server will send. The ASGI server is not responsible for validating
+the ``Trailers`` headers provided.
 
 Keys:
 
@@ -165,6 +169,13 @@ Keys:
   ``[name, value]`` two-item iterables, where ``name`` is the header name, and
   ``value`` is the header value. Header names must be lowercased. Pseudo
   headers (present in HTTP/2 and HTTP/3) must not be present.
+
+* ``more_trailers`` (*bool*): Signifies if there is additional content
+  to come (as part of a *HTTP Trailers* message). If ``False``, response
+  will be taken as complete and closed, and any further messages on
+  the channel will be ignored. Optional; if missing defaults to
+  ``False``.
+
 
 The ASGI server will only send the trailing headers in case the client has sent the
 ``TE: trailers`` header in the request.
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -133,3 +133,38 @@ TLS
 ---
 
 See :doc:`specs/tls`.
+
+HTTP Trailers
+---
+
+The Trailer response header allows the sender to include additional fields at the
+end of chunked messages in order to supply metadata that might be dynamically
+generated while the message body is sent, such as a message integrity check,
+digital signature, or post-processing status.
+
+ASGI servers that implement this extension will provide
+``http.response.trailers`` in the extensions part of the scope::
+
+    "scope": {
+        ...
+        "extensions": {
+            "http.response.trailers": {},
+        },
+    }
+
+An ASGI framework interested in sending trailing headers to the client, must include
+the ``Trailer`` header in the ``http.response.start``. That will allow the ASGI server
+to know that after the last ``http.response.body`` message (``more_body`` being ``False``),
+the ASGI framework will send a ``http.response.trailers`` message.
+
+Keys:
+
+* ``type`` (*Unicode string*): ``"http.response.trailers"``
+
+* ``headers`` (*Iterable[[byte string, byte string]]*): An iterable of
+  ``[name, value]`` two-item iterables, where ``name`` is the header name, and
+  ``value`` is the header value. Header names must be lowercased. Pseudo
+  headers (present in HTTP/2 and HTTP/3) must not be present.
+
+The ASGI server will only send the trailing headers in case the client has sent the
+``TE: trailers`` header in the request.
--- a/specs/www.rst
+++ b/specs/www.rst
@@ -185,14 +185,19 @@ Keys:
   lowercased. Optional; if missing defaults to an empty list. Pseudo
   headers (present in HTTP/2 and HTTP/3) must not be present.
 
+* ``trailers`` (*bool*) -- Signifies if the application will send
+  trailers. If ``True``, the server must wait until it receives a
+  ``"http.response.trailers"`` message after the *Response Body* event.
+  Optional; if missing defaults to ``False``.
+
 
 Response Body - ``send`` event
 ''''''''''''''''''''''''''''''
 
 Continues sending a response to the client. Protocol servers must
 flush any data passed to them into the send buffer before returning from a
-send call. If ``more_body`` is set to ``False`` this will
-close the connection.
+send call. If ``more_body`` is set to ``False``, and the server is not
+expecting *Response Trailers* this will close the connection.
 
 Keys:
 
@@ -203,10 +208,10 @@ Keys:
   missing defaults to ``b""``.
 
 * ``more_body`` (*bool*) -- Signifies if there is additional content
-  to come (as part of a Response Body message). If ``False``, response
-  will be taken as complete and closed, and any further messages on
-  the channel will be ignored. Optional; if missing defaults to
-  ``False``.
+  to come (as part of a *Response Body* message). If ``False``, and the
+  server is not expecting *Response Trailers* response will be taken as
+  complete and closed, and any further messages on the channel will be
+  ignored. Optional; if missing defaults to ``False``.
 
 
 Disconnect - ``receive`` event


docs/extensions.rst
```diff
@@ -133,3 +133,38 @@ TLS
...
+
+* ``type`` (*Unicode string*): ``"http.response.trailers"``
+
+* ``headers`` (*Iterable[[byte string, byte string]]*): An iterable of
+  ``[name, value]`` two-item iterables, where ``name`` is the header name, and
+  ``value`` is the header value. Header names must be lowercased. Pseudo
+  headers (present in HTTP/2 and HTTP/3) must not be present.
+
+The ASGI server will only send the trailing headers in case the client has sent the
+``TE: trailers`` header in the request.
```
pgjones, line 170: Should the server error on receipt of this message without the `TE: trailers` header or just silently not send the trailing headers?

docs/extensions.rst
```diff
@@ -133,3 +133,38 @@ TLS
...
+
+    "scope": {
+        ...
+        "extensions": {
+            "http.response.trailers": {},
+        },
+    }
+
+An ASGI framework interested in sending trailing headers to the client, must include
+the ``Trailer`` header in the ``http.response.start``. That will allow the ASGI server
```
pgjones, line 156: Is the `Trailer` header the same as `TE: trailers` mentioned below? Or should there be an additional field?

Kludex: No, this `Trailer` header is a response header. The `TE: trailers` below is from the request header.

Example of full interaction:
- The client sends a `TE: trailers` request header.
- The server sends `Trailer: X-Server-Timing` in the response headers.
- Body is sent.
- `X-Server-Timing` is sent as trailing headers.

docs/extensions.rst
Kludex: I think it should be more on the "silently not send" side.

docs/extensions.rst
synodriver, line 156: Is the ``trailer`` field in ``http.response.start`` necessary for server implementations？I didn't have that field but nonecorn still works.

docs/extensions.rst
synodriver: > Should the server error on receipt of this message without the `TE: trailers` header or just silently not send the trailing headers?

In my implementation, the protocol server just send trailers when it receive that event. I think it's the responsibility of the asgi application to decide whether to send or not. So I don't have ```trailer``` field in ```http.response.start```. If you want to send, just send.

docs/extensions.rst
Kludex: How do you make sure that the server will not close the transport without informing that the trailing headers will be available? Making the _last_ `http.response.body` to send `more_body=True`? 🤔

synodriver: >  Making the last http.response.body to send more_body=True

Yes, this is exactly what I am doing.

> I didn't add the trailer field in the http.response.start on this spec. I've mentioned that the Trailer header should be included in the headers field on the http.response.start.

I saw it from [uvicorn's pr](https://github.com/encode/uvicorn/pull/1599/files) which seems to have a trailer field.

Kludex: > I saw it from [uvicorn's pr](https://github.com/encode/uvicorn/pull/1599/files) which seems to have a trailer field.

I think you didn't see it, but I mentioned on the description of this PR:

> The uvicorn one introduces a new key on the `http.response.start` called `trailers`, to show the intention from the application to the server that trailing headers will be sent. I think this is not needed. The server can understand the intention if the header `Trailing:` is provided on the `headers` from the `http.response.start`.

Kludex: > Yes, this is exactly what I am doing. And add a filed for that event, like { 'type': 'http.response.trailers', 'trailers': [ [b'x-trailer-test', b'test'], ], "more_body": False }

Got it. But then... The ASGI framework still needs to send the `Trailer` header. So the `more_body` set to `False` on the last `http.response.body` is also not needed, because the server already knows about the `Trailer` header.

synodriver: Actually I didn't look into the headers when I wrote that... So you mean we can improve it by letting server automatically judge from the headers? I just let asgi applications send as they want, no need to add extra headers.

synodriver: Well, according to the RFC, your implementation is more accurate. I will fix that soon. But what will happen if an asgi application wants to send trailers multiple times?

Kludex: According to https://www.rfc-editor.org/rfc/rfc2616.html#section-14.40:

>  An HTTP/1.1 message SHOULD include a Trailer header field in a
   message using chunked transfer-coding with a non-empty trailer. Doing
   so allows the recipient to know which header fields to expect in the
   trailer.

So it should be included... So I guess, we need to make the server know at the beginning of if it's going to send trailing headers.

Kludex: > Well, according to the RFC, your implementation is more accurate. I will fix that soon. But what will happen if an asgi application wants to send trailers multiple times?

By this spec, it's not allowed. 🤔 Should it be? By this spec, I'd expect a single `http.response.trailers` message with all the headers in the `headers` (or we can call it `trailers` if wanted) field.

synodriver: I think more than one ```http.response.trailers``` should be possible for some cases, and it's not very hard for server  implementations.

Kludex: I was writing the changes that I would have to implement if this is the case... So I wrote this:

> If I adjust the spec to allow that, then I'll have to add the following changes in the document:
> - The last `http.response.body` will send `more_body` as `True`.
> - Add `more_body` field to `http.response.trailers`. 
> - The last `http.response.trailers` will send `more_body` as `False`.
> Just to be clear, the ASGI application will still be in charge of sending the `Trailer:` header.

But then I noticed... What if the `Trailer:` header is not sent by the ASGI application? Does the ASGI server waits for all `http.response.trailers`, and does not send those to the client? Does the server closes the transport and then raises an error if `http.response.trailers` are detected without the initial `Trailer:` header?

synodriver: Does the ASGI application need to ensure that the data is sent as required, or does the server have the obligation to check?

synodriver: If the check is left to the ASGI app, the implementation of the server will be much easier, and we can build a prototype server from here. After all, the server is called by the ASGI application. If the ASGI application runs fine, there will be no problem. The server only needs to check the necessary fields for sending data.

Kludex: > The server only needs to check the necessary fields for sending data.

The server SHOULD let the client knows the trailing headers it will be sending, as I've mentioned above:

> An HTTP/1.1 message SHOULD include a Trailer header field in a
message using chunked transfer-coding with a non-empty trailer. Doing
so allows the recipient to know which header fields to expect in the
trailer.

I think you're right, the server should not care if the trailing headers keys were sent in the `http.response.start`. But... The ASGI application SHOULD send them.

About the "multiple `http.response.trailers` messages": If that's allowed, then we need a field to allow the server to understand it has more. I guess `more_body` it's the candidate field here...

docs/extensions.rst
```diff
@@ -133,3 +133,38 @@ TLS
...
+to know that after the last ``http.response.body`` message (``more_body`` being ``False``),
+the ASGI framework will send a ``http.response.trailers`` message.
```
Kludex, line 158:
```suggestion
An ASGI framework interested in sending trailing headers should include the ``Trailer`` header
in the *Response Start*, as mentioned in the [RFC 2616](https://www.rfc-editor.org/rfc/rfc2616.html#section-14.40).

The ASGI framework that wants to send trailing headers can send multiple *Response Body* messages,
but the last one should set ``more_body`` as ``True``. This will tell the ASGI server to wait for
more messages, on which the ASGI framework can send multiple ``http.response.trailers`` messages.
The last ``http.response.trailers`` on its turn should set ``more_body`` as ``False``.
```

docs/extensions.rst
```diff
@@ -133,3 +133,38 @@ TLS
...
+
+Keys:
+
+* ``type`` (*Unicode string*): ``"http.response.trailers"``
+
+* ``headers`` (*Iterable[[byte string, byte string]]*): An iterable of
+  ``[name, value]`` two-item iterables, where ``name`` is the header name, and
+  ``value`` is the header value. Header names must be lowercased. Pseudo
+  headers (present in HTTP/2 and HTTP/3) must not be present.
```
Kludex, line 167:
```suggestion
  headers (present in HTTP/2 and HTTP/3) must not be present.
  
* ``more_body`` (*bool*): Signifies if there is additional content
  to come (as part of a Response Body message). If ``False``, response
  will be taken as complete and closed, and any further messages on
  the channel will be ignored. Optional; if missing defaults to
  ``False``.
```

Kludex (commented):
@synodriver According to what we talked, is this what you think it should be written?

docs/extensions.rst
synodriver: The ```more_body```  can be ``False`` by default if not exists, just follow the one in ```http.response.body```

synodriver: So multiple ``http.response.trailers`` messages is allowed, but if you just want to send one, you may not to add ``more_body`` field. This is what ``http.response.body`` did.

docs/extensions.rst
synodriver: Yes, exactly.

docs/extensions.rst
Kludex: Yeah, that field is False by default, as mentioned below.

docs/extensions.rst
synodriver: I have changed nonecorn to adapt this spec, and with a [simple test app](https://gist.github.com/synodriver/66abc632f42f75e6b05ef60aa3a484d8). I tested it with curl and looks fine.

Kludex: Thanks @synodriver 🙏 Let's see what others have to say here. If everything is fine, I'll commit those 2 suggestions above.

docs/extensions.rst
pgjones: I think that multiple `http.response.trailers` messages is an unnecessary complication - headers are unlikely to be streamed and I don't believe it makes sense to send more body after the trailers. Therefore I'd leave it for the ASGI app to figure out what trailers to send and have a single message sent to the server.

Kludex: @pgjones Then the server needs to be aware that the trailing headers are going to be present on `http.response.start`, and `more_body` is not needed on `http.response.trailers` (what is currently commited).

Then are we fine on saying that the ASGI framework MUST send the `Trailer` header in the `http.response.start`?

pgjones: I think so - does the ASGI app need to give a value equal to the headers it intends to send as well?

synodriver: > I think that multiple http.response.trailers messages is an unnecessary complication 

Actually when you are using [h11](https://pypi.org/project/h11/), there is only one chance to send trailer: the end of response body.  So it's ok for servers which only support http/1.1. But when it comes to http/2, things changed. Mutiple trailers is allowed, and, to be honest, I spend quite a lot time thinking trailers under http/2, and found that may need some [tricky](https://github.com/nonebot/nonecorn/issues/11). I have no idea if it's allowed.

Kludex: > I think so - does the ASGI app need to give a value equal to the headers it intends to send as well?

In case the ASGI app doesn't send the `Trailer` header, how does the server can be aware when the response is completed? 

On the `uvicorn` PR mentioned on the description, it uses a new field `trailers` on the `http.response.start`. On the `nonecorn` implementation, it makes use of the `more_body` on `http.response.body` and `http.response.trailers`.

pgjones: I agree with the ASGI app only needing to include a `Trailer` header, my question was if the server should check the values of that header against the trailers sent.

Kludex: Ah! I would naively say "no". As the RFC mentioned above says, the server SHOULD (and not MUST) send the `Trailer` header, I don't think it's necessary for the server to make sure the trailings headers are the same as the ones in the `Trailer` header.

Should I clarify this in the spec?

Kludex:
Moving @synodriver's comment to its own thread:

> > I think that multiple http.response.trailers messages is an unnecessary complication
> 
> Actually when you are using [h11](https://pypi.org/project/h11/), there is only one chance to send trailer: at the end of response body. So it's ok for servers which only support http/1.1. But when it comes to http/2, things changed. Mutiple trailers is allowed, and, to be honest, I spend quite a lot time thinking trailers under http/2, and found that may need some [tricky](https://github.com/nonebot/nonecorn/issues/11). I have no idea if it's allowed.

I'm not sure how this has implications in the spec, but I'd like to highlight this comment, as it may get lost.

docs/extensions.rst
```diff
@@ -133,3 +133,38 @@ TLS
 ---
 
 See :doc:`specs/tls`.
+
```
Kludex, line 136: Moving @synodriver's comment to its own thread:

> > I think that multiple http.response.trailers messages is an unnecessary complication
> 
> Actually when you are using [h11](https://pypi.org/project/h11/), there is only one chance to send trailer: at the end of response body. So it's ok for servers which only support http/1.1. But when it comes to http/2, things changed. Mutiple trailers is allowed, and, to be honest, I spend quite a lot time thinking trailers under http/2, and found that may need some [tricky](https://github.com/nonebot/nonecorn/issues/11). I have no idea if it's allowed.

I'm not sure how this has implications in the spec, but I'd like to highlight this comment, as it may get lost.

tomchristie:
(Because I've been cc'ed)

Does it matter if we've seen enough of a genuine use case or not? I'm aware that HTTP trailing headers exist in the spec, but I've never encountered their usage in my career. So I'd be a bit wary of introducing that complexity into the spec, and the expectation that can then introduce onto other projects to include that complexity in their implementations.

I'd probably put an "HTTP trailing headers" extension into the "technology for the sake of technology", unless there was a clearer indication that it's solving someone's actual use-case.

Perhaps that'd involve something like a "here's a functional gRPC implementation that we can show works if we have this extension and doesn't work without it" (?)

But, also shrug. 🤷‍♂️

"Sensible pushback" or "Mildly grumpy old geezer"? You decide. 😇

sam-kleiner:
> (Because I've been cc'ed)
> 
> Does it matter if we've seen enough of a genuine use case or not? I'm aware that HTTP trailing headers exist in the spec, but I've never encountered their usage in my career. So I'd be a bit wary of introducing that complexity into the spec, and the expectation that can then introduce onto other projects to include that complexity in their implementations.
> 
> I'd probably put an "HTTP trailing headers" extension into the "technology for the sake of technology", unless there was a clearer indication that it's solving someone's actual use-case.
> 
> Perhaps that'd involve something like a "here's a functional gRPC implementation that we can show works if we have this extension and doesn't work without it" (?)
> 
> But, also shrug. 🤷‍♂️
> 
> "Sensible pushback" or "Mildly grumpy old geezer"? You decide. 😇

I did the (uvicorn PR)[https://github.com/encode/uvicorn/pull/1599] and didnt realize this would be so controversial. 

I have an actual need for this and some use cases are laid out in that PR. But in short, It is useful when streaming a response in a couple cases. You may encounter an error that you want the client to know about, as you have already responded 200 and sent headers how else could you get this information to them. You may also be streaming a large file that you want to compute the hash of as it stream through the sever and send that hash at the end.

sam-kleiner:
Copying some comments from my separate PR above.

I wanted to open this to add options to the discussion. I did the (uvicorn implementation)[https://github.com/https://github.com/encode/uvicorn/pull/1599]. Personally I like the more_body being set to false when the body is done and trailers being sent after. It makes more sense to me, but at the end of the day if any trailers get added I will be happy.

docs/extensions.rst
sam-kleiner: I did it the same way in my separate PR. I agree the responsibility should be on the asgi app

andrewgodwin:
I think I slightly prefer the design proposed in #341 - it's more explicit, and doesn't rely on parsing the content of headers sent back. Is there any strong reason not to go with that design? (I would want to add the extension signifier over there too, rather than just have it appear in the spec and force a new version)

Kludex:
So by slightly prefer you mean the `trailers` field being added to the `http.response.start` instead of the server inferring? Or is there another difference?

I'm fine with this. The client will still should to send what trailing headers are those, but it's not mandatory.

Is the `more_trailers` needed?

andrewgodwin:
Yes, to be specific:
* Trailers field in the start message is better than inferring
* more_trailers should be there in case the server is chunking them (because if someone is already using trailers, the chances they're doing strange things is much higher)

Kludex:
Is there a scenario on which we have this sequence:
1. `http.response.start`
2. `http.response.trailers`
Without the `http.response.body`?

Kludex:
I used the name `headers` instead of `trailers`, as there's already a `trailers` field being introduced in *_Response Start_*, and all the `headers` field have the same type... But I can change that if wanted.

asgiref/typing.py
```diff
@@ -99,6 +100,7 @@ class HTTPResponseStartEvent(TypedDict):
     type: Literal["http.response.start"]
     status: int
     headers: Iterable[Tuple[bytes, bytes]]
+    trailers: bool
```
andrewgodwin, line 103: TypedDict is `total` by default, so this would require `trailers` to always be present which it won't be (as we're doing this as a non-major change by making it optional). We need to switch this so that `trailers` is optional, which may need to be done with two inheriting types due to TypedDict not having fine-grained options.

andrewgodwin (commented):
I think I'm OK with it being called `headers` in the trailer response - that is, after all, what they are, rather than trailing body.

asgiref/typing.py
Kludex: `more_body` is `bool` on `HTTPResponseBodyEvent`. What's the difference between those?

abersheeran: > We need to switch this so that `trailers` is optional, which may need to be done with two inheriting types due to TypedDict not having fine-grained options.

We can use https://peps.python.org/pep-0655/#specification

Kludex: What both are suggesting was already discussed when the `asgiref.typing` module was implemented. See https://github.com/django/asgiref/pull/221#issuecomment-751387856.

andrewgodwin: It was, but that was "types for the first time" whereas this is modifying existing types - if we land this change, anything that currently (correctly) typechecks as a response no longer will, and that seems bad.

The suggestion in that thread of Unioning two other dict types together would also work, though.

Kludex: > It was, but that was "types for the first time"

What's the difference? `asgiref` is versioned... :thinking:

Kludex: >  if we land this change, anything that currently (correctly) typechecks as a response no longer will, and that seems bad.

Take in consideration that `uvicorn` is the only ASGI server implementation that uses `asgiref.typing`... :man_shrugging:

andrewgodwin (approved):
OK, I think this is good enough to go. I'll leave a bit more time for any further objections and then we'll merge it.

Kludex:
Thanks everybody for the comments! :pray:

PR #341, opened 2022-08-17
RE: Add HTTP Trailers 

sam-kleiner:
There is already a [PR](https://github.com/django/asgiref/pull/339) open regarding this. I wanted to open this to add options to the discussion. I did the [uvicorn implementation](https://github.com/encode/uvicorn/pull/1599).

As stated in the other proposal PR this works a bit differently. This adds a new message type `http.response.trailers` and a new `trailers` bool property to `http.response.start`. IMO this simplifies the implementation. 

My original POC did not account for chunked trailers but this would be easy to add with a `more_trailers` property similar to the `more_body` property.

Personally I like the `more_body` being set to false when the body is done and trailers being sent after. It makes more sense to me, but at the end of the day if any trailers get added I will be happy.

### DIFF
--- a/asgiref/typing.py
+++ b/asgiref/typing.py
@@ -16,6 +16,7 @@ __all__ = (
     "HTTPRequestEvent",
     "HTTPResponseStartEvent",
     "HTTPResponseBodyEvent",
+    "HTTPResponseTrailersEvent",
     "HTTPServerPushEvent",
     "HTTPDisconnectEvent",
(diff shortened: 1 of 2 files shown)


andrewgodwin:
I do like this in general, but I feel like we need a `more_trailers` in there because people are inevitably going to want that?

Kludex:
This can be closed. It was already merged on https://github.com/django/asgiref/pull/339.

PR #342, opened 2022-08-23
Early hints extension

pgjones:
This gives an ASGI mechanism for the application to indicate to the
server what early hints to send.

See https://developers.cloudflare.com/cache/about/early-hints/, https://httpwg.org/specs/rfc8297.html, and https://www.rfc-editor.org/rfc/rfc8288.html.

I hope to implement this in Hypercorn and replace my server push usage.

### DIFF
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -133,3 +133,37 @@ TLS
 ---
 
 See :doc:`specs/tls`.
+
+Early Hints
+-----------
+
+An informational response with the status code 103 is an Early Hint,
+indicating to the client that resources are associated with the
+subsequent response, see ``RFC 8297``. ASGI servers that implement
+this extension will allow early hints to be sent. These servers will
+provide ``http.response.early_hint`` in the extensions part of the
+scope::
+
+    "scope": {
+        ...
+        "extensions": {
+            "http.response.early_hint": {},
+        },
+    }
+
+An ASGI framework can send an early hint by sending a message with the
+following keys. This message can be sent at any time (and multiple
+times) after the *Response Start* message but before the final
+*Response Body* message.
+
+Keys:
+
+* ``type`` (*Unicode string*): ``"http.response.early_hint"``
+
+* ``links`` (*Iterable[byte string]*): An iterable of link header field
+  values, see ``RFC 8288``.
+
+The ASGI server should then attempt to send an informational response
+to the client with the provided links as ``Link`` headers. The server
+may decide to ignore this message, for example if the HTTP/1.1
+protocol is used and the server has security concerns.


docs/extensions.rst
```diff
@@ -133,3 +133,37 @@ TLS
 ---
 
 See :doc:`specs/tls`.
+
+Early hints
+-----------
+
+A informational response with the status code 103 is an early hint,
```
Kludex, line 140:
```suggestion
An informational response with the status code 103 is an early hint,
```

docs/extensions.rst
```diff
@@ -133,3 +133,37 @@ TLS
...
+*Response Body* message.
+
+Keys:
+
+* ``type`` (*Unicode string*): ``"http.response.early_hint"``
+
+* ``links`` (*Iterable[byte string]*): An iterable of link header field
+  values, see ``RFC 8288``.
+
+The ASGI server should then attempt to send an infromational response
```
Kludex, line 166:
```suggestion
The ASGI server should then attempt to send an informational response
```

docs/extensions.rst
pgjones: Thanks, fixed

docs/extensions.rst
pgjones: Thanks fixed

andrewgodwin:
Ah yes, with Server Push basically being officially dead it does seem like time. This looks pretty good - any concerns or other ways you might want to do it you avoided here?

pgjones:
I had thought of a more general `http.informational_response` message, but I can't think of why the ASGI app should want to send any other types. I also considered being more prescriptive about the link structure, but I think that is something the ASGI app can structure and decide like other headers.

docs/extensions.rst
```diff
@@ -133,3 +133,37 @@ TLS
 ---
 
 See :doc:`specs/tls`.
+
+Early hints
```
andrewgodwin, line 137: Let's capitalise Hints here, as that's how everyone refers to it

docs/extensions.rst
```diff
@@ -133,3 +133,37 @@ TLS
...
+-----------
+
+An informational response with the status code 103 is an early hint,
```
andrewgodwin, line 140: Can we have Early Hint either capitalised or in quotes here to indicate it is the proper name of the HTTP feature?

andrewgodwin (commented):
OK, sounds sensible - a couple pf tiny copy-edit changes and then I'm pretty happy with this.

docs/extensions.rst
pgjones: Yep, done

docs/extensions.rst
pgjones: Yep, done

andrewgodwin:
Lovely, thanks. Let's hope this gains much more traction overall than Server Push!

PR #350, opened 2022-11-06
Update lifespan.rst

Kludex:


andrewgodwin:
For backwards compatibility, I believe this needs to stay the same. All 2.0 versions should provide it.

Kludex:
I think this commit 4 years ago missed this update: https://github.com/django/asgiref/commit/55e9ac45f15e89df58b1e9af54e5fe17be8f9322

Isn't it the case?

andrewgodwin:
That's for the `www` spec, which is different to `lifespan` in terms of versioning.

Kludex:
I see... Thanks.

PR #351, opened 2022-11-06
Add `trailers` field to `WebSocketResponseStartEvent`

Kludex:
image


andrewgodwin:
There are no trailers specified in the spec for Websocket events, so we can't add this.

andrewgodwin:
I don't believe you can even send trailers on WebSocket connections sensibly, since I don't know how the upgrade response would take it. Either way, I don't want to support it - if you want to put in a PR that fixes the wording to say that websockets have everything from the normal response _except_ for trailers, then go for it.

PR #352, opened 2022-11-06
Remove `None` parameter from `websocket.close`

Kludex:
The reasoning behind this is to avoid confusion. None of the other parameters on this document have this "None" conditional included...


andrewgodwin:
I believe there was a reason for it at the time, and as a spec, we can't make backwards incompatible changes like this without a major revision.

PR #354, opened 2022-11-11
Add extensions["lifespan.state"]

### LINKED ISSUE #322: Persistence of state from lifespan events to requests
ASGI lifespan events are designed to handle setup and subsequent teardown of resource like a database connection, but does not provide any persistence to store this state to requests.

Starlette implements an `Application.state` namespace ([docs](https://www.starlette.io/applications/#storing-state-on-the-app-instance)) which modifies the application object in place. Quart suggest to store data in the app's namespace directly.
This is not ideal because it gives an otherwise stateless thing state, and there is also no correlation between this state and the event loop / ASGI state, which can lead to inconsistent state.
Here's an artificial but not absurd example of how this could lead to a confusing user experience:

```python
import asyncio
from contextlib import asynccontextmanager
from dataclasses import dataclass
from typing import AsyncIterator

from starlette.applications import Starlette
from starlette.testclient import TestClient
from starlette.requests import Request
from starlette.responses import Response
from starlette.routing import Route


@dataclass
class Connection:
    loop: asyncio.AbstractEventLoop

    async def run(self) -> None:
        if asyncio.get_event_loop() is not self.loop:
            raise Exception("Some obtuse error")


@asynccontextmanager
async def lifespan(app: Starlette) -> AsyncIterator[None]:
    app.state.db = Connection(asyncio.get_event_loop())
    yield


async def endpoint(request: Request) -> Response:
    await request.app.state.db.run()
    return Response()


app = Starlette(
    routes=[Route("/", endpoint)],
    lifespan=lifespan,
)


def some_test_using_lifespans() -> None:
    with TestClient(app):
        pass


def some_test_where_the_user_forgets_to_use_the_lifespan() -> None:
    TestClient(app).get("/")


if __name__ == "__main__":
    some_test_using_lifespans()
    some_test_where_the_user_forgets_to_use_the_lifespan()
```

Here an honest mistake ends up leaking state between tests, and maybe giving the user an obtuse error about event loops and such.

I think it would be beneficial if the spec provided a namespace scoped to each ASGI lifespan / request. This namespace would basically be an empty dict that gets persisted from the lifespan event into each request. I think it would make sense to model this like contextvar propagation into tasks: every request gets a _copy_ of the namespace from the lifespan (unless there was no lifespan, in which it's a brand new dict or maybe None).
Then state can be naturally stored in this namespace.
Since the ASGI server manages this namespace (and it already manages it's own state because of lifespans), the application can be stateless and neither the framework nor the users have to worry about clearing state at the end of a lifespan or anything like.

This can easily be backwards compatible: it's a new key in the scope and we bump the minor version of the ASGI version so that frameworks can check if this is supported or not.

adriangb:
Closes #322

### DIFF
--- a/asgiref/typing.py
+++ b/asgiref/typing.py
@@ -101,7 +101,7 @@ class WebSocketScope(TypedDict):
 class LifespanScope(TypedDict):
     type: Literal["lifespan"]
     asgi: ASGIVersions
-    state: Optional[Dict[str, Any]]
+    state: NotRequired[Dict[str, Any]]
 
 
 WWWScope = Union[HTTPScope, WebSocketScope]
--- a/asgiref/typing.py
+++ b/asgiref/typing.py
@@ -1,10 +1,15 @@
 import sys
-from typing import Awaitable, Callable, Dict, Iterable, Optional, Tuple, Type, Union
+from typing import Any, Awaitable, Callable, Dict, Iterable, Optional, Tuple, Type, Union
 
 if sys.version_info >= (3, 8):
     from typing import Literal, Protocol, TypedDict
 else:
     from typing_extensions import Literal, Protocol, TypedDict
+
+if sys.version_info >= (3, 11):
+    from typing import NotRequired
+else:
+    from typing_extensions import NotRequired
 
 __all__ = (
     "ASGIVersions",
@@ -62,6 +67,7 @@ class HTTPScope(TypedDict):
     headers: Iterable[Tuple[bytes, bytes]]
     client: Optional[Tuple[str, int]]
     server: Optional[Tuple[str, Optional[int]]]
+    state: NotRequired[Dict[str, Any]]
     extensions: Optional[Dict[str, Dict[object, object]]]
 
 
@@ -78,12 +84,14 @@ class WebSocketScope(TypedDict):
     client: Optional[Tuple[str, int]]
     server: Optional[Tuple[str, Optional[int]]]
     subprotocols: Iterable[str]
+    state: NotRequired[Dict[str, Any]]
     extensions: Optional[Dict[str, Dict[object, object]]]
 
 
 class LifespanScope(TypedDict):
     type: Literal["lifespan"]
     asgi: ASGIVersions
+    state: Optional[Dict[str, Any]]
 
 
 WWWScope = Union[HTTPScope, WebSocketScope]
--- a/asgiref/typing.py
+++ b/asgiref/typing.py
@@ -1,5 +1,15 @@
 import sys
-from typing import Any, Awaitable, Callable, Dict, Iterable, Optional, Tuple, Type, Union
+from typing import (
+    Any,
+    Awaitable,
+    Callable,
+    Dict,
+    Iterable,
+    Optional,
+    Tuple,
+    Type,
+    Union,
+)
 
 if sys.version_info >= (3, 8):
     from typing import Literal, Protocol, TypedDict
--- a/specs/lifespan.rst
+++ b/specs/lifespan.rst
@@ -46,6 +46,9 @@ The scope information passed in ``scope`` contains basic metadata:
 * ``asgi["version"]`` (*Unicode string*) -- The version of the ASGI spec.
 * ``asgi["spec_version"]`` (*Unicode string*) -- The version of this spec being
   used. Optional; if missing defaults to ``"1.0"``.
+* ``state`` Optional(*dict[Unicode string, Any]*) -- An empty namespace where
+  the application can persist state to be used when handling subsequent requests.
+  Optional; if missing the server does not support this feature.
 
 If an exception is raised when calling the application callable with a
 ``lifespan.startup`` message or a ``scope`` with type ``lifespan``,
@@ -56,6 +59,38 @@ lifespan protocol. If you want to log an error that occurs during lifespan
 startup and prevent the server from starting, then send back
 ``lifespan.startup.failed`` instead.
 
+The ``extensions["lifespan.state"]`` dict is an empty namespace.
+Applications can store arbitrary data in this namespace.
+A *shallow copy* of this dictionary will get passed into each request handler.
+This key will only be set if the server supports this extension.
+
+
+Lifespan State
+--------------
+
+Applications often want to persist data from the lifespan cycle to request/response handling.
+For example, a database connection can be established in the lifespan cycle and persisted to
+the request/response cycle.
+The ```lifespan["state"]`` namespace provides a place to store these sorts of things.
+The server will ensure that a *shallow copy* of the namespace is passed into each subsequent
+request/response call into the application.
+Since the server manages the application lifespan and often the event loop as well this
+ensures that the application is always accessing the database connection (or other stored object)
+that corresponds to the right event loop and lifecycle, without using context variables,
+global mutable state or having to worry about references to stale/closed connections.
+
+ASGI servers that implement this feature will provide
+``state`` as part of the ``lifespan`` scope::
+
+    "scope": {
+        ...
+        "state": {},
+    }
+
+The namespace is controlled completely by the ASGI application, the server will not
+interact with it other than to copy it.
+Nonetheless applications should be cooperative by properly naming their keys such that they
+will not collide with other frameworks or middleware.
 
 Startup - ``receive`` event
 '''''''''''''''''''''''''''
--- a/specs/www.rst
+++ b/specs/www.rst
@@ -121,6 +121,10 @@ The *connection scope* information passed in ``scope`` contains:
   listening port, or ``[path, None]`` where ``path`` is that of the
   unix socket. Optional; if missing defaults to ``None``.
 
+* ``state`` Optional(*dict[Unicode string, Any]*) -- A copy of the
+  namespace passed into the lifespan corresponding to this request. (See :doc:`lifespan`).
+  Optional; if missing the server does not support this feature.
+
 Servers are responsible for handling inbound and outbound chunked transfer
 encodings. A request with a ``chunked`` encoded body should be automatically
 de-chunked by the server and presented to the application as plain body bytes;
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,7 @@ python_requires = >=3.7
 packages = find:
 include_package_data = true
 install_requires =
-    typing_extensions; python_version < "3.8"
+    typing_extensions; python_version < "3.11"
 zip_safe = false
 
 [options.extras_require]


specs/lifespan.rst
```diff
@@ -46,6 +46,8 @@ The scope information passed in ``scope`` contains basic metadata:
 * ``asgi["version"]`` (*Unicode string*) -- The version of the ASGI spec.
 * ``asgi["spec_version"]`` (*Unicode string*) -- The version of this spec being
   used. Optional; if missing defaults to ``"1.0"``.
+* ``extensions["state"]`` Optional(*dict[Unicode string, Any]*) -- An empty namespace where
+  the application can persist state to be used when handling subsequent requests.
```
Kludex, line 50: Isn't this supposed to be only "state", or did we agree on this "extensions" field previously?

```suggestion
* ``state`` Optional(*dict[Unicode string, Any]*) -- An empty namespace where
  the application can persist state to be used when handling subsequent requests.
```

specs/www.rst
```diff
@@ -121,12 +121,8 @@ The *connection scope* information passed in ``scope`` contains:
   listening port, or ``[path, None]`` where ``path`` is that of the
   unix socket. Optional; if missing defaults to ``None``.
 
-Servers are responsible for handling inbound and outbound chunked transfer
-encodings. A request with a ``chunked`` encoded body should be automatically
-de-chunked by the server and presented to the application as plain body bytes;
-a response that is given to the server with no ``Content-Length`` may be chunked
-as the server sees fit.
```
Kludex, line 128: Why was this removed? :thinking:

Kludex:
The types module needs changes as well, jfyk.

specs/lifespan.rst
adriangb: The current spec says that extensions can be used when "a change to a specification is being trialled before being rolled in". That said, I'd be happy to just slap it into the main namespace. It would be a backwards compatible change, i.e. Starlette would just transparently use it instead of it's own dict:
https://github.com/encode/starlette/blob/eb76552d53215de7a319335903dd8a0aebd2df7f/starlette/requests.py#L167-L169. It sounds like your vote would be for main namespace?

specs/www.rst
adriangb: Mistake, thank you for pointing it out 😄

specs/lifespan.rst
Kludex: Oh, no! I'm not in favor of breaking Starlette hahaha

I'm not sure if I get all you said, but what I mean is I don't see the "extensions" namespace being used besides here... :thinking:

adriangb: > I'm not in favor of breaking Starlette hahaha

My point was we would _not_ be breaking Starlette! In fact, using the existing `"state"` key could be a good thing because it means _less_ changes in Starlette to fully support this and some support out of the box as soon as Uvicorn is updated.

> I don't see the "extensions" namespace being used besides here

Do you mean in this section of the docs or in the spec in general? We definitely use it, e.g. `extensions["http.response.trailers"]`.

Kludex:
While looking at this PR I thought: "Why not introducing a new field on `lifespan.startup.complete`?" I had this thought because it would be the first extension that would change the scope on the ASGI application side. 

The thing is, if we do that, the scope on the `shutdown` event needs to have the `state` field anyway, because we want to have those resources available there. In case the application sends the `state` on the `lifespan.startup.complete`, the scope on the `lifespan.startup`, and on the `lifespan.shutdown` will have to differ on this field.

---

I'm going to ping some people that might be interested in this. :wave: @pgjones @tomchristie

asgiref/typing.py
```diff
@@ -1,11 +1,16 @@
...
 
 if sys.version_info >= (3, 8):
     from typing import Literal, Protocol, TypedDict
 else:
     from typing_extensions import Literal, Protocol, TypedDict
 
+if sys.version_info >= (3, 11):
+    from typing import NotRequired
+else:
+    from typing_extensions import NotRequired
```
Kludex, line 12: I don't think the NotRequired should be included only here. This was previously mentioned on the HTTP trailers PR, and on the PR that introduced this module (I'm on my phone, difficult to provide links).

IMO either we change what we consider to be optional on those type hints, or we don't add the NotRequired.

Kludex (commented):
There were some rebases here, so we lost a bit the track of the previous comments. I'm going to give a bit of context.

The first implementation here was adding the "Lifespan State" extension to the "Extensions" section under the `extensions["lifespan.state"]`, but it was also adding the `extensions["state"]` on the lifespan scope. But there's no precedent about using the "extensions" key outside the "Extensions" section. So... The only thing missing on that first implementation was to change the `extensions["state"]` to `state` - ofc being optional.

The current implementation, I believe is a breaking change, based on previous comments from Andrew. Since it adds a subsection under the "Lifespan" section.

Anyway, just expressing here what we talked outside GitHub.

adriangb:
Thank you for all of the clarifying comments Marcelo.

> There's no precedent about using the "extensions" key outside the "Extensions" section.

I don't think this is a particularly salient reason to make it an extension or not. My original reasoning for making it an extension was to avoid conflict with Starlette's use of `scope["state"]`. I changed it because I realized that putting a dict in `scope["state"]` would not be a breaking change for Starlette and in fact would work out really nicely because Starlette would already get some support out of the box.

> So... The only thing missing on that first implementation was to change the extensions["state"] to state - ofc being optional.

I didn't really understand this comment.

That said I'm happy to move it to `extensions`, swap `NotRequired` for `Optional`; I don't think any of these change the core proposal so I don't have strong feelings.

Kludex:
> I don't think this is a particularly salient reason to make it an extension or not. My original reasoning for making it an extension was to avoid conflict with Starlette's use of scope["state"]. I changed it because I realized that putting a dict in scope["state"] would not be a breaking change for Starlette and in fact would work out really nicely because Starlette would already get some support out of the box.

I don't think we are understanding each other. Let me try to improve my wording:
1. I **do** think this should be an extension.
2. I **do** think it should be called `extensions["lifespan.state"]`.
3. I **do not** think the namespace to store the state should be `extensions["state"]` - it should be only `state`.

Complying with the 3rd point doesn't make this proposal a non-extension.

Analogous, let's take the HTTP Trailers previously implemented... The ASGI server needs to provide a `extensions["http.response.trailers"]` in the scope, and the *Response Start* needs to provide the `trailers` field, as mentioned [here](https://asgi.readthedocs.io/en/latest/specs/www.html#response-start-send-event). The fact that the *Response Start* now has an optional `trailers` field doesn't make this proposal a non-extension one.

---

As for the fact that Starlette is using this field... I think the most correct thing would be to find a namespace that doesn't conflict with a namespace that a known ASGI framework is already using, unless we are trying to document what the ASGI framework is already doing. Because in any case, if we do this, Starlette users are going to receive a different data on that namespace, so it's a breaking change.

Maybe we can find a name that has the same meaning as "state"? - Let's read Andrew's comment before...

:point_down: Unrelated to this PR:
- To avoid this kind of confusion in the future, I do believe that we need to go back to the discussion about a namespace for ASGI frameworks to do what they want, and not just accept "any that is unlikely to have conflicts in the future". :shrug:

adriangb:
> I don't think we are understanding each other. Let me try to improve my wording:
> 
> 1. I **do** think this should be an extension.
> 2. I **do** think it should be called `extensions["lifespan.state"]`.
> 3. I **do not** think the namespace to store the state should be `extensions["state"]` - it should be only `state`.
> 
> Complying with the 3rd point doesn't make this proposal a non-extension.

Thanks for clarifying, that makes more sense. That sounds fine to me! I'll make the changes to reflect this.

> As for the fact that Starlette is using this field... I think the most correct thing would be to find a namespace that doesn't conflict with a namespace that a known ASGI framework is already using, unless we are trying to document what the ASGI framework is already doing. Because in any case, if we do this, Starlette users are going to receive a different data on that namespace, so it's a breaking change.

Why would it be a breaking change? User's code will only break if they were doing something like `request.state.foo = getattr(request.state, "foo", 123)` and also use this new feature to set `scope["extensions"]["lifespan.state"]["foo"] = 456`. I don't think anyone is doing that because (1) why would they check if the field is set or not that dict was always empty previously and (2) they'd have to be opting into this feature *and* clobbering their own namespace.

I think re-using the field that is already being used is a good thing, it means Starlette and things that depend on it or imitate it (so most of the ASGI landscape) can start using this feature without making any changes internally. For example, I could add support to https://github.com/adriangb/asgi-lifespan while we figure out the API in Starlette to actually set values in the lifespan and then those values would show up transparently in the request/response cycle's `scope["state"]` and thus Starlette's `Request.state`.

Kludex:
> Thanks for clarifying, that makes more sense. That sounds fine to me! Since we've gone back and forth I'll wait for Andrew to chime in though before making any changes. I do want to point out that "lifespan.state" seems a bit redundant since it's inside the lifespan scope, what other state could it be?

The extension needs to be provided in the HTTP scope as well, but I think it's fine using the `extensions["state"]` anyway.

> Why would it be a breaking change? User's code will only break if they were doing something like request.state.foo = getattr(request.state, "foo", 123) and also use this new feature to set scope["extensions"]["lifespan.state"]["foo"] = 456. I don't think anyone is doing that because (1) why would they check if the field is set or not that dict was always empty previously and (2) they'd have to be opting into this feature and clobbering their own namespace.

Oh, yeah, actually, the user needs to add the field itself on the state initially. Got it. So yeah, the "state" field is fine. :+1:

Kludex:
@andrewgodwin anything we can do to push this forward?

andrewgodwin:
Sorry, I've been very busy with other things for the past month or so - I'll hopefully get to this soonish, but it's going to take time for me to re-read the initial ticket here to be in the right headspace.

adriangb:
@andrewgodwin if you are too busy to review that's okay, but would you be able to recommend or be okay with tagging another reviewer? This PR/issue have been waiting for an in-depth review for almost a year now.

asgiref/typing.py
adriangb: The thing is that making it `Optional` implies that the server _must_ set it to None, so every single server is immediately out of spec. Using `NotRequired` means that no changes are necessary to servers.

andrewgodwin:
We don't exactly have anyone else who reviews things on this repo, sadly. I would love another maintainer! But they don't exactly appear out of thin air.

I've just got back from Australia; I'll try to sit down and get to this this week.

andrewgodwin (approved):

andrewgodwin:
Alright, I have re-read it, understand it, and it looks good. Fix the linting error and I'll land it!

asgiref/typing.py
```diff
@@ -78,12 +94,14 @@ class WebSocketScope(TypedDict):
...
     server: Optional[Tuple[str, Optional[int]]]
     subprotocols: Iterable[str]
+    state: NotRequired[Dict[str, Any]]
     extensions: Optional[Dict[str, Dict[object, object]]]
 
 
 class LifespanScope(TypedDict):
     type: Literal["lifespan"]
     asgi: ASGIVersions
+    state: NotRequired[Dict[str, Any]]
```
adriangb, line 104: @andrewgodwin gave this a last pass and noticed that this was `Optional[Dict[str, Any]]` but I think it should be `NotRequired[Dict[str, Any]]` for the same reasons as https://github.com/django/asgiref/pull/354/files#r1118125962 so I changed it. Just an FYI since you already reviewed.

andrewgodwin: I agree - I also missed that!

andrewgodwin:
Thanks for your work (and persistence) on this!

adriangb:
Thank you for reviewing and pushing this across the finish line! I'm excited to explore the new APIs this will unblock.

I think I should have changed the PR title since the commit message is now out of date with the implementation, sorry about that...

specs/www.rst
```diff
@@ -121,6 +121,10 @@ The *connection scope* information passed in ``scope`` contains:
   listening port, or ``[path, None]`` where ``path`` is that of the
   unix socket. Optional; if missing defaults to ``None``.
 
+* ``state`` Optional(*dict[Unicode string, Any]*) -- A copy of the
+  namespace passed into the lifespan corresponding to this request. (See :doc:`lifespan`).
+  Optional; if missing the server does not support this feature.
+
```
Kludex, line 127: This is missing in the `WebSocket` section below.

specs/lifespan.rst
```diff
@@ -56,6 +59,38 @@ lifespan protocol. If you want to log an error that occurs during lifespan
 startup and prevent the server from starting, then send back
 ``lifespan.startup.failed`` instead.
 
+The ``extensions["lifespan.state"]`` dict is an empty namespace.
+Applications can store arbitrary data in this namespace.
```
Kludex, line 63: This line is unclear... In the next section it's specified to store in `lifespan["state"]`, but here it mentions that applications can use `extension["lifespan.state"]`... :thinking:

Was this supposed to be removed?

adriangb: Yes I think so

asgiref/typing.py
HansBrende: @adriangb why can't this be `Dict[Any, Any]` rather than `Dict[str, Any]`?

Below, the docs state: 
> Applications can store arbitrary data in this namespace.

Why wouldn't that include arbitrary keys?

> applications should be cooperative by properly naming their keys such that they
will not collide with other frameworks or middleware.

The most foul-proof way to avoid collisions would be to not use string keys at all, but objects (similar to `Symbol` keys in javascript).

Also, that would help make dependency injection simpler in FastAPI/Starlette.
Example use-case:

```python
# (roughly)
HTTP_CLIENT: ClientSession = fastapi.Depends(lambda request: request.scope['state'][HTTP_CLIENT])
DB_POOL: Pool = fastapi.Depends(lambda request: request.scope['state'][DB_POOL])

@asynccontextmanager
async def app_lifespan(app: FastAPI):
    async with make_http_client() as client:
        async with make_database_connection_pool() as db_pool:
            yield {HTTP_CLIENT: client, DB_POOL: db_pool}
```

I am already doing this and it works just fine! However, it is annoying that the typing doesn't match up quite right... and I don't see a reason keys should be limited to strings.

adriangb: Starlette already assumes string keys for its state object, so I kind of copied that here.
It's always possible to do `scope["state"]["mykey"] = {object(): 123}`.
Finally, I think if FastAPI ends up wanting to use this feature and shows a compelling use case for making the keys `Any` we can re-evaluate and change the typing.

HansBrende: @adriangb thanks for the response. This isn't something I'd think would be part of the fastapi codebase (although, maybe, who knows). This is more a user-land example of a use-case for non-string keys. Would it be more appropriate for me to raise a similar issue in Starlette or Uvicorn then? As I mentioned, I am already doing the above approach (nothing prevents me). It's just that the typing is a lie, since I'm storing non-string keys. Since Starlette don't have any control over what kinds of keys are returned from the app lifespan function (example above), and similarly, Uvicorn doesn't have any control over what keys its lifespan state dict is updated with, it doesn't make sense to me to pretend they are definitely strings.

adriangb: One thing in particular that will break is Starlette's `State` class: https://github.com/encode/starlette/blob/15d9350eed9437bcb0683e2c24cd4cfbf985ed7e/starlette/datastructures.py#L683-L708. Now, I'm not a fan of that class/functionality, but it exists. Of course, if you don't care you can continue doing what you are doing, but then solving the typing issue is also as simple as a `type: ignore`.
I'm also curious why the suggestion above of something like `yield {"dependencies": {...}}` or something like that wouldn't work.

All of that said, if you feel strongly please feel free to open an issue / PR so it can be discussed properly.

PR #365, opened 2022-12-27
Add Debug extension

Kludex:
- Closes https://github.com/django/asgiref/issues/135

### DIFF
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -221,7 +221,7 @@ The debug extension allows a way to send debug information from an ASGI framewor
 its responses. This extension is not meant to be used in production, only for testing purposes,
 and ASGI servers should not implement it.
 
-The ASGI context sent to the application will provide ``http.response.debug`` in the extensions
+The ASGI context sent to the framework will provide ``http.response.debug`` in the extensions
 part of the scope::
 
     "scope": {
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -217,9 +217,9 @@ The ASGI server will only send the trailing headers in case the client has sent 
 Debug
 -----
 
-The debug extension allows a way to send debug information from the ASGI application to
-an entity acting as an ASGI server. This extension is not meant to be used in production,
-but only for testing purposes, and ASGI servers should not implement it.
+The debug extension allows a way to send debug information from an ASGI application in
+its responses. This extension is not meant to be used in production, only for testing purposes,
+and ASGI servers should not implement it.
 
 The entity acting as ASGI server will provide ``http.response.debug`` in the extensions
 part of the scope::
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -213,3 +213,32 @@ Keys:
 
 The ASGI server will only send the trailing headers in case the client has sent the
 ``TE: trailers`` header in the request.
+
+Debug
+-----
+
+The debug extension allows a way to send debug information from the ASGI application to
+an entity acting as an ASGI server. This extension is not meant to be used in production,
+but only for testing purposes, and ASGI servers should not implement it.
+
+The entity acting as ASGI server will provide ``http.response.debug`` in the extensions
+part of the scope::
+
+    "scope": {
+        ...
+        "extensions": {
+            "http.response.debug": {},
+        },
+    }
+
+The ASGI framework can send debug information by sending a message with the following
+keys. This message can be sent at any time (and multiple times), even before the *Response
+Start* message.
+
+Keys:
+
+* ``type`` (*Unicode string*): ``"http.response.debug"``
+
+* ``info`` (*Dict[Unicode string, Any]*): A dictionary containing the debug information.
+  The keys and values of this dictionary are not defined by the ASGI specification, and
+  are left to the ASGI framework to define.
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -217,7 +217,7 @@ The ASGI server will only send the trailing headers in case the client has sent 
 Debug
 -----
 
-The debug extension allows a way to send debug information from an ASGI application in
+The debug extension allows a way to send debug information from an ASGI framework in
 its responses. This extension is not meant to be used in production, only for testing purposes,
 and ASGI servers should not implement it.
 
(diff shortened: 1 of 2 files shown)


docs/extensions.rst
```diff
@@ -213,3 +213,32 @@ Keys:
 
 The ASGI server will only send the trailing headers in case the client has sent the
 ``TE: trailers`` header in the request.
+
+Debug
+-----
+
+The debug extension allows a way to send debug information from the ASGI application to
+an entity acting as an ASGI server. This extension is not meant to be used in production,
```
Kludex, line 221: Can we improve the "entity acting as an ASGI server"?

docs/extensions.rst
```diff
@@ -213,3 +213,32 @@ Keys:
...
+but only for testing purposes, and ASGI servers should not implement it.
+
+The entity acting as ASGI server will provide ``http.response.debug`` in the extensions
```
Kludex, line 224: I've added the `response` here, but I'm not sure if I should have...

The suggestions were `http.debug_info` and `http.test_info`, but I find redundant to add the word "info", since everything is info here... :eyes: I can change it tho...

docs/extensions.rst
```diff
@@ -213,3 +213,32 @@ Keys:
...
+    "scope": {
+        ...
+        "extensions": {
+            "http.response.debug": {},
+        },
+    }
+
+The ASGI framework can send debug information by sending a message with the following
+keys. This message can be sent at any time (and multiple times), even before the *Response
+Start* message.
```
Kludex, line 236: I think this "any time" may complicate things... Starlette sends it before the *Response Start*, should I be strict and comply on what Starlette does here?

docs/extensions.rst
andrewgodwin: I'd just drop it and just say "from an ASGI application in its responses".

docs/extensions.rst
andrewgodwin: No, I think the new name is fine. I'd also remove "entity..." here and switch to "the ASGI context sent to the application"

docs/extensions.rst
```diff
@@ -213,3 +213,32 @@ Keys:
 
 The ASGI server will only send the trailing headers in case the client has sent the
 ``TE: trailers`` header in the request.
+
+Debug
+-----
+
+The debug extension allows a way to send debug information from the ASGI application to
+an entity acting as an ASGI server. This extension is not meant to be used in production,
+but only for testing purposes, and ASGI servers should not implement it.
```
Kludex, line 222:
```suggestion
The debug extension allows a way to send debug information from an ASGI application in
its responses. This extension is not meant to be used in production, only for testing purposes,
and ASGI servers should not implement it.
```

docs/extensions.rst
Kludex: Like this? :thinking: 
```suggestion
The ASGI context sent to the application will provide ``http.response.debug`` in the extensions
```

Kludex:
Just as a note... I think the issue that motivated this PR was more than just a "debug" extension.

I think it was more about having a namespace on which the application could add any kind of data related to the execution - because right now the ASGI applications use any field that may not conflict with the ASGI spec itself. I'm mentioned this based on the whole issue, but specifically about [this comment](https://github.com/django/asgiref/issues/135#issuecomment-554553318).

In any case, maybe it's interesting to discuss if we should really add this "debug" extension or just add document a namespace to "execution" information. :shrug:

andrewgodwin:
My recollection of the original motivation (granted, it's been 3 years) was that it was purely for testing - scenarios where you would be passing the raw ASGI response into more code. Server behaviour should be governed by their own extensions (private or spec'd), I think.

docs/extensions.rst
andrewgodwin: Yup!

docs/extensions.rst
andrewgodwin: Yeah, let's be a bit more specific because it can't hurt, and say it precedes response start. There's only very few scenarios where you'd want it later.

Kludex:
Yeah. This PR already satisfies my needs on Starlette (which is documented on the issue).

docs/extensions.rst
Kludex: Which scenarios do you see "later"?

andrewgodwin: All I could think of was things like "how many sql queries we did while streaming a response out". I was tempted to say we should always make it the _last_ message, but that then doesn't let you test endpoints that stream indefinitely.

Kludex: I think that can already be done by the HTTP trailers extension, if someone wants to do it. I guess...

docs/extensions.rst
```diff
@@ -213,3 +213,32 @@ Keys:
...
+    "scope": {
+        ...
+        "extensions": {
+            "http.response.debug": {},
+        },
+    }
+
+The ASGI framework can send debug information by sending a message with the following
+keys. This message can be sent at any time (and multiple times), even before the *Response
+Start* message.
```
Kludex, line 236:
```suggestion
keys. This message must be sent once, before the *Response Start* message.
```

docs/extensions.rst
Kludex: Changed.

docs/extensions.rst
```diff
@@ -213,3 +213,31 @@ Keys:
...
+part of the scope::
+
+    "scope": {
+        ...
+        "extensions": {
+            "http.response.debug": {},
+        },
+    }
+
+The ASGI framework can send debug information by sending a message with the following
```
andrewgodwin, line 234: Nit: this says "framework" where above it says "application". They should probably be consistent.

andrewgodwin (commented):
Looks good apart from a single word!

docs/extensions.rst
```diff
@@ -213,3 +213,31 @@ Keys:
 
 The ASGI server will only send the trailing headers in case the client has sent the
 ``TE: trailers`` header in the request.
+
+Debug
+-----
+
+The debug extension allows a way to send debug information from an ASGI application in
```
Kludex, line 220:
```suggestion
The debug extension allows a way to send debug information from an ASGI framework in
```

Kludex:
Applied 🙏

andrewgodwin (approved):

andrewgodwin:
Lovely, thanks for your work on this!

Issue #346, opened 2022-09-30
Example apps in docs


adamchainz:
The spec currently defines a lot of details without any example apps (except in the lifespan section). It would be good to include basic examples, perhaps function and class versions for both full apps and middleware.

Kludex:
Do we want one for each: HTTP, WebSockets, and extensions?

If I receive the Andrew's ack here, I can work on a PR. :+1:

andrewgodwin:
If you can somehow make them clean and simple, sure!

Kludex:
Is this clean and simple: https://github.com/django/asgiref/pull/366 ?

andrewgodwin:
It is!

PR #366, opened 2022-12-28
Add HTTP protocol example

Kludex:
- Relates to #346


andrewgodwin:
Could we get it in a separate example section?

Kludex:
Hmm... Like a new "Examples" section?

I did like this because it's analogous to the "Lifespan" section.

andrewgodwin:
Yes, a new Examples section close to the bottom of the doc. I just want to make sure this upper section is kept relatively the same.

emarhin (approved):

Issue #270, opened 2021-06-14
Preserving typying on `async_to_sync`/`sync_to_async` wrapped functions


shughes-uk:
Currently wrapping functions with these class based decorators will obliterate type information. With the introduction of `Paramspec` in python 3.10 and the backports library `typing_extensions` it should be possible (using `Generic`) to preserve this information.

The code for this looks like this

```python
from typing_extensions import ParamSpec
from typing import Callable, Generic, TypeVar

return_value  = TypeVar("return_value")
parameters = ParamSpec("parameters")

class FooDecorator(Generic[parameters, return_value]):
    def __init__(self,func: Callable[parameters, return_value]):
        self.func = func

    def __call__(self,*args:parameters.args, **kwargs:parameters.kwargs)-> return_value:
        return self.func(*args,**kwargs)
```

This would currently be blocked by a bug in the typing backport (https://github.com/python/typing/pull/817), but is soon to resolved. I would be happy to create a PR for this once the bug is resolved.

andrewgodwin:
Yes, let's get this in once it's possible with the backport. The decorators have always been very annoying when typing is involved!

shughes-uk:
The blocking MR has now been released as part of the 3.10 version of typing extensions. Any blockers to me having a go at implementing this?

andrewgodwin:
I don't think so, I'd love someone else to take on mypy stuff rather than having to stare at it for hours!

graingert:
mypy 0.920 is out now

LucidDan:
Not realising there was an existing older PR in #298 that I could've based off, I've added a new PR that addresses this in #373...however, it appears we took mostly the same approach. 

Done plenty of typing on private projects, but this is a first for an open source project, so feedback would be appreciated.

LucidDan:
Also worth noting and discussing; the minimum mypy version for this PR to work is 0.991. Given mypy 1.0.0+ is out now, I'm not sure how much of an issue this is anymore, but it still might be deserving of some consideration as to whether it will cause people issues if they are forced to upgrade their mypy version. At the least, it probably means this should only be added in a minor asgiref release (eg 3.7.0) ?

LucidDan:
This got resolved in #390

Issue #372, opened 2023-02-18
Installing pre-commit hooks for asgiref causes a (confusing!) "poetry" related error


LucidDan:
I've been doing some work on a typing PR for asgiref, and ran into a problem installing pre-commit hooks:

```
(asgiref) ➜  asgiref git:(main) ✗ pre-commit install-hooks
[INFO] Installing environment for https://github.com/pycqa/isort.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: command: ('/Users/dans/.cache/pre-commit/repoon7vdpun/py_env-python3.11/bin/python', '-mpip', 'install', '.')
return code: 1
stdout:
    Processing /Users/dans/.cache/pre-commit/repoon7vdpun
      Installing build dependencies: started
      Installing build dependencies: finished with status 'done'
      Getting requirements to build wheel: started
      Getting requirements to build wheel: finished with status 'done'
      Preparing metadata (pyproject.toml): started
      Preparing metadata (pyproject.toml): finished with status 'error'
    
stderr:
      error: subprocess-exited-with-error
      
      × Preparing metadata (pyproject.toml) did not run successfully.
      │ exit code: 1
      ╰─> [17 lines of output]
          Traceback (most recent call last):
            File "/Users/dans/.cache/pre-commit/repoon7vdpun/py_env-python3.11/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
              main()
            File "/Users/dans/.cache/pre-commit/repoon7vdpun/py_env-python3.11/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 335, in main
              json_out['return_val'] = hook(**hook_input['kwargs'])
                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
            File "/Users/dans/.cache/pre-commit/repoon7vdpun/py_env-python3.11/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 149, in prepare_metadata_for_build_wheel
              return hook(metadata_directory, config_settings)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #373, opened 2023-02-19
Add ParamSpec typing for AsyncToSync and SyncToAsync

LucidDan:
This uses typing-extensions to add ParamSpec typing to the decorators and their related classes sync_to_async and async_to_sync.

The typing-extensions dependency needed to be extended to python 3.9, to provide backported support for ParamSpec.

Also found through some manual testing that the minimum mypy version that works with ParamSpec is 0.991. Maybe this means this PR needs to wait for at least an asgiref minor release, as it will affect the version of mypy people need to be using.

### DIFF
--- a/asgiref/current_thread_executor.py
+++ b/asgiref/current_thread_executor.py
@@ -1,6 +1,9 @@
 import queue
 import threading
 from concurrent.futures import Executor, Future
+from typing import Any, TypeVar
+
+T = TypeVar("T")
 
 
 class _WorkItem:
@@ -35,12 +38,14 @@ class CurrentThreadExecutor(Executor):
     the thread they came from.
     """
 
-    def __init__(self):
+    _work_queue: "queue.Queue[Any]"
+
+    def __init__(self) -> None:
         self._work_thread = threading.current_thread()
         self._work_queue = queue.Queue()
         self._broken = False
 
-    def run_until_future(self, future):
+    def run_until_future(self, future: "Future[T]") -> None:
         """
         Runs the code in the work queue until a result is available from the future.
         Should be run from the thread the executor is initialised in.
--- a/asgiref/sync.py
+++ b/asgiref/sync.py
@@ -9,13 +9,34 @@ import threading
 import warnings
 import weakref
 from concurrent.futures import Future, ThreadPoolExecutor
-from typing import Any, Callable, Dict, Optional, overload
+from types import TracebackType
+from typing import (
+    Any,
+    Awaitable,
+    Callable,
+    Dict,
+    Generic,
+    List,
+    Optional,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
+    cast,
+    overload,
+)
+
+# Use typing_extensions until python 3.9 support is dropped
+from typing_extensions import ParamSpec
 
 from .current_thread_executor import CurrentThreadExecutor
 from .local import Local
 
+T = TypeVar("T")
+P = ParamSpec("P")
 
-def _restore_context(context):
+
+def _restore_context(context: contextvars.Context) -> None:
     # Check for changes in contextvars, and set them to the current
     # context for downstream consumers
     for cvar in context:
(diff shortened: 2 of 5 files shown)


LucidDan:
After doing the work, I realised there was at least one existing draft PR in #298 , albeit a couple of years old now. The approach we took seems pretty similar; I've included at least one change to my PR from the feedback in that PR's comments.

LucidDan:
Worth noting; this probably won't pass all tests until #372 is fixed.

LucidDan:
Feedback welcomed, I haven't done ParamSpec generics on a publicly used project before...I am not 100% sure if this will play well with users' deployments, but I've tested it with my own projects and seems okay so far.

Also, to link it into the original issue - this is to fix #270

markedwards:
One comment that I think applies to both of these implementations — `Coroutine[Any, Any, R]` is better than `Awaitable[R]` for `sync_to_async`, because it will raise an error if the wrapped function is not awaited.  This is not an error for an `Awaitable`, and that can lead to ugly bugs.

I’ve been using this to work around the lack of typing and it works well:

```
class SyncToAsyncCallable(Protocol):
    def __call__(
        self,
        func: Callable[P, R],
        thread_sensitive: bool = True,
        executor: Optional[ThreadPoolExecutor] = None,
    ) -> Callable[P, Coroutine[Any, Any, R]]: ...


sync_to_async: Final[SyncToAsyncCallable] = asgiref.sync.sync_to_async
```

LucidDan:
> One comment that I think applies to both of these implementations — `Coroutine[Any, Any, R]` is better than `Awaitable[R]` for `sync_to_async`, because it will raise an error if the wrapped function is not awaited. This is not an error for an `Awaitable`, and that can lead to ugly bugs.

Thanks for the suggestion Mark!
I've only used `Callable[P, Awaitable[T]]` in `AsyncToSync`, where I believe it makes sense, because you can absolutely pass a callable to it that returns some other type of awaitable object (for example, a Future, instead of a Coroutine object). For that case, I don't think it makes sense to narrow the type more than necessary, but I'm happy to get further input on that if I've missed something.

For `sync_to_async`/`SyncToAsync`, there's no Awaitable or Coroutine typing in my PR; there's no need, because the __call__ is an "async def", which makes it implicitly a Coroutine return type, so that's already covered.

Again, let me know if I'm missing something or misunderstood you here. Feel free to annotate the PR changes with specific comments if you want, as well.

markedwards:
> For `sync_to_async`/`SyncToAsync`, there's no Awaitable or Coroutine typing in my PR; there's no need, because the `__call__` is an "async def", which makes it implicitly a Coroutine return type, so that's already covered.

Got it, that makes sense.  Obviously I didn’t look closely enough at your diff yet.  I’ll comment directly on it if I see anything.

andrewgodwin (commented):
Agreed with the idea that this will need to be a minor release due to the mypy dep change.

I've looked over it and it all seems to make sense to me; I've not dabbled with ParamSpec more than a few times, but the arguments being passed through without modification makes this simpler than the times I have.

My one question - is there a way to write a test for this? Obviously a standard unit test is not going to work, but something that fails the typechecker if it's not correct might be possible.

LucidDan:
> My one question - is there a way to write a test for this? Obviously a standard unit test is not going to work, but something that fails the typechecker if it's not correct might be possible.

Good question; I'd thought about this in passing but now you've prompted me to go looking- upon a little research found that Sobolevn wrote about it a while back and mentions complex generics specifically - https://sobolevn.me/2019/08/testing-mypy-types
Short tldr version - there is a pytest plugin that allows testing for expected error cases with multiple. I'll experiment with it and see if I can add some tests to the PR.

LucidDan:
To sum up my findings so far:
- pytest-mypy-plugins is useful; I wish it worked in a python module style rather than YML definitions, but either way, it's a good fit for testing types on a python package. I think it could be quite useful in general, not just for this PR.
- I've added a test spec (tests/test_sync_types.yml) to the PR that currently fails. I found that mypy fails to correctly pass through the paramspec and typevar through the SyncToAsync and AsyncToSync classes; I get errors indicating that the return type is 'Any'.
- When I tried to reproduce with a minimal test case (eg https://gist.github.com/mypy-play/f7005753d50af85ba6a9c7242291c448), it doesn't fail. It also doesn't fail in pylance, for what that's worth.

I'll take a further look in the next couple of days as time allows and try to figure out what the issue is.

And sorry, I might've made a mess of the review history due to a force-push or two while I was trying to wrangle the commit history.

Kludex:
We do this on Starlette: https://github.com/encode/starlette/blob/master/tests/test_config.py#L12-L30

LucidDan:
> We do this on Starlette: https://github.com/encode/starlette/blob/master/tests/test_config.py#L12-L30

Thanks, that does avoid the extra dependency at least, I'll explore that as an alternative too.

PR #379, opened 2023-03-12
Fix typo in lifespan state spec

florimondmanca:
Hi there,

This is a fix for the "Lifespan state" spec docs: https://asgi.readthedocs.io/en/latest/specs/lifespan.html#lifespan-state

The prose mentions `lifespan["state"]` but the example code shows `scope["state"]`. My understanding is that the latter is the correct reference.

cc @Kludex


Kludex (approved):
:+1:

andrewgodwin (approved):

PR #385, opened 2023-04-20
Allow Local instances to be cleared without deleting

ababic:
Greetings, lovely people! 

I'm in the process of using `Local` in a couple of projects, and recently had a case where I wanted to clear   cached values based on certain signals, without deleting the object completely. 

I'm sure you'll agree the changes are super simple, but I feel this tiny change brings the API more in-line with other familiar cache options, and I think may help with adoption.

Thank you for all the wonderful work so far, folks.

### DIFF
--- a/asgiref/local.py
+++ b/asgiref/local.py
@@ -84,7 +84,7 @@ class Local:
             self._context_refs.add(context_obj)
         return getattr(context_obj, self._attr_name)
 
-    def __del__(self):
+    def clear(self):
         try:
             for context_obj in self._context_refs:
                 try:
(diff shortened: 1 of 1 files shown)


andrewgodwin (approved):

andrewgodwin:
I like the idea, but like anything that touches Local, it looks like the tests need a bit more attention!

ababic:
Thanks @andrewgodwin. I was previously relying on the CI to see test results, but I'll get things setup here properly and ensure tests are all ship-shape before requesting a review.

PR #386, opened 2023-04-27
Better garbage-collection on exceptions in sync_to_async

tahajahangir:
This fixes a garbage-collection issue explained in: https://stackoverflow.com/q/76118191/326792

Current implementation of `sync_to_async` blocks python from garbage-collecting failed tasks with reference-counting. The explicit assignment of local variable `source_task` to `None` fixes this!

### DIFF
--- a/asgiref/sync.py
+++ b/asgiref/sync.py
@@ -481,6 +481,9 @@ class SyncToAsync:
         else:
             self.launch_map[current_thread] = source_task
             parent_set = True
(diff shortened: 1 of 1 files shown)


andrewgodwin (approved):

andrewgodwin:
If you could fix the one linting issue, this looks good and I'd be happy to merge it!

andrewgodwin:
Thanks!

PR #387, opened 2023-05-02
Update www.rst

Kludex:
The other day I noticed those words in the spec, and neither Uvicorn or the Starlette TestClient are following this "wait for body" constraints.


andrewgodwin (approved):

andrewgodwin:
Yes, it was indeed inherited from WSGI! I think it's fine to remove it given it's not followed and the world hasn't collapsed :)

Kludex:
Thanks Andrew 🙏

kristjanvalur:
Well, that clears that up :)

PR #390, opened 2023-05-12
Improve sync.py typing and utilities performance

bellini666:
* Make sure `sync_to_async`/`async_to_sync` keeps the original function typing signature

* Define iscoroutinefunction/markcoroutinefunction at import time instead of an if/else in runtime.

The current implementation:
```python
In [7]: timeit.timeit(lambda: iscoroutinefunction(some_sync_func), number=10000)
Out[7]: 0.012024379917420447
```

Versus this PR
```python
In [11]: timeit.timeit(lambda: iscoroutinefunction(some_sync_func), number=10000)
Out[11]: 0.004663120023906231
```

This is a performance improvement of almost 300%.

### DIFF
--- a/asgiref/current_thread_executor.py
+++ b/asgiref/current_thread_executor.py
@@ -2,7 +2,7 @@ import queue
 import sys
 import threading
 from concurrent.futures import Executor, Future
-from typing import Any, Callable, TypeVar, Union
+from typing import TYPE_CHECKING, Any, Callable, TypeVar, Union
 
 if sys.version_info >= (3, 10):
     from typing import ParamSpec
@@ -22,7 +22,7 @@ class _WorkItem:
 
     def __init__(
         self,
-        future: Future[_R],
+        future: "Future[_R]",
         fn: Callable[_P, _R],
         *args: _P.args,
         **kwargs: _P.kwargs,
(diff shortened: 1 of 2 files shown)


andrewgodwin (commented):
Code changes look reasonable, but you will need to please the mypy in the CI to get this in!

bellini666:
@andrewgodwin thanks for the review. Just solved all the issues to please mypy :)

andrewgodwin:
Unfortunately not for 3.9 and below, it looks like - the problem with ParamSpec is it's relatively new, and you'll need a way of keeping things backward-compatible somehow.

I don't care much about 3.7 as it's about to be EOL, but 3.8 and 3.9 do still matter.

(Apologies I can't tell our CI to run for you on every commit!)

bellini666:
> Unfortunately not for 3.9 and below, it looks like - the problem with ParamSpec is it's relatively new, and you'll need a way of keeping things backward-compatible somehow.
> 
> I don't care much about 3.7 as it's about to be EOL, but 3.8 and 3.9 do still matter.

Oh, sorry about that. The code at `sync.py` was already conditionally importing from typing only for 3.10+, but I forgot to do the same when fixing typing on `current_thread_executer.py`

I only have python 3.11 in my local dev environment, so I'll only know for older versions once the CI gets to run

> (Apologies I can't tell our CI to run for you on every commit!)

Np :)

Hopefully everything should be fine now

andrewgodwin:
Still something awry it looks like. You should be able to run the full set of python versions tests locally with `tox` if you'd like.

bellini666:
> Still something awry it looks like. You should be able to run the full set of python versions tests locally with tox if you'd like.

@andrewgodwin I was being lazy to install all required python versions =P. Just did that, ran tox and was able to _really_ fix everything now.

Sorry about that!

andrewgodwin:
Yup, all looks good now! Thanks, I appreciate you taking the time to go through all those revisions!

PR #396, opened 2023-05-25
refactor: type sync_to_async/async_to_async as returning callables instead of the SyncToAsync/AsyncToSync

### LINKED ISSUE #394: Argument missing for parameter
There's a bug with the type hinting in version 3.7.0+

```python
from asgiref.sync import sync_to_async


class Whatever:
    @sync_to_async()
    def something(self, i: int) -> int:
        return i + 1


what = Whatever()


async def main() -> None:
    await what.something(i=1)  # error is here
```

Produces the following error in `pyright`:

> Argument missing for parameter "self" [reportGeneralTypeIssues]

And with `mypy`:

> error: Too few arguments  [call-arg]

This code did not produce any type hinting issues in version 3.6.0 and prior.

- pyright 1.1.309
- mypy 1.30
- python 3.10.10

bellini666:
Fix #394

### DIFF
--- a/asgiref/sync.py
+++ b/asgiref/sync.py
@@ -554,8 +554,54 @@ class SyncToAsync(Generic[_P, _R]):
             return None
 
 
-# Lowercase aliases (and decorator friendliness)
-async_to_sync = AsyncToSync
+@overload
+def async_to_sync(
+    *,
+    force_new_loop: bool = False,
+) -> Callable[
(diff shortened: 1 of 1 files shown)


andrewgodwin (approved):

andrewgodwin:
Thanks - I have released this as 3.7.2 just now.

Issue #399, opened 2023-06-13
Using `contextvars` in `local.py`?


akx:
I noticed `local.py` has a comment saying

https://github.com/django/asgiref/blob/a90aeb93c424c471400118879cb652cddc1726a8/asgiref/local.py#L29-L30

which is at odds with the changelog entry for 3.7.0 of this library:

https://github.com/django/asgiref/blob/a90aeb93c424c471400118879cb652cddc1726a8/CHANGELOG.txt#L14-L18

(I noticed `sync.py` uses `contextvars.Context`, but `local.py` doesn't. Maybe there was confusion there?)

Either way, what would be the "nicer storage"? Is someone working on it? :)

andrewgodwin:
Yeah, we switched sync to use it, but not Local yet. I don't think anyone is actively working on it.

bluetech:
I looked at this briefly some time ago when looking at `Local`'s performance. IIRC (and I might be misremembering), the difficulty was that `contextvars` is rooted in a thread local, while `Local` can be inherited across threads. The semantics are quite different.

spanezz:
Unless I'm missing something, this seems to be done in https://github.com/django/asgiref/commit/d920c3c44f59ef037be3a09a41a5640014cfb1e2

andrewgodwin:
It was indeed done - I will close this as a result, we'll discuss the ramifications elsewhere.

PR #401, opened 2023-07-07
Adjusted changelog for 3.7.0 to reflect that `Local` isn't `contextvars` yet

akx:
As per https://github.com/django/asgiref/issues/399#issuecomment-1589687227


andrewgodwin:
:+1:

PR #402, opened 2023-08-02
Clarify that raw_path does not include the query string

simonw:
This was a point of confusion here:
- https://github.com/encode/httpx/issues/2810


andrewgodwin:
Thanks!

PR #403, opened 2023-08-14
Disable Thread sensitive

wasimTem:
Merln using channels 2.2.0 and thread sensitive off is the current fastest option


wasimTem:
f

andrewgodwin:
This would be a totally backwards-incompatible change so we can't so it - thread_sensitive mode is slightly slower but slightly safer. If you know it's safe to turn it off without, for example, corrupting database connections, you are welcome to do that in your own codebase!

wasimTem:
bro im sorry i meant to pr to my own repo misclick

PR #407, opened 2023-09-05
Add `pathsend` extension

### LINKED ISSUE #404: Proposal: introduce `http.response.pathsend` extension
Some ASGI servers might not be able to implement `http.response.zerocopysend` extension due to specific implementation constraints, for instance Rust implemented servers like [Granian](https://github.com/emmett-framework/granian) won't play well with file descriptors.

Nevertheless, such servers might still be able to directly serve files with "zero copy" techniques when the source path is known in advance. For instance, Granian already provide such feature within the [RSGI protocol](https://github.com/emmett-framework/granian/blob/master/docs/spec/RSGI.md#http-protocol-interface), offering ~28x performance increase over standard file serving in ASGI.

Given this dramatic performance gain, it would be nice to introduce a similar approach also in ASGI, applying the same *offloading* technique for such kind of contents, where the server itself is the sole responsible of processing the files. And this is the aim of this proposal: introduce an extension quite similar to `http.response.zerocopysend` but based on paths rather than file descriptiors.

The idea is to add a `pathsend` extension which the server can list in the appropriate key of the scope:

```python
"scope": {
    ...
    "extensions": {
        "http.response.pathsend": {},
    },
}
```

With the above extra definition, a new event can be produced by ASGI applications, with the following keys:

- `type` (Unicode string): "http.response.pathsend"
- `pathsend` (Unicode string): The string representation of the file path to serve (platform specific)

which will instruct the server to serve the file directly from the relevant path.

A couple of details worth to mention:
- in this design, the ASGI application is still responsible to send the `http.response.start` event with the relevant headers (eg: `content-type`, `content-length`, etc.)
- the path should be absolute, or when relative would be relative to the current working directory the server was started

Once we discussed this properly, and if it makes sense, I'm up to open a PR with the relevant changes.
Reference of the Granian issue: https://github.com/emmett-framework/granian/issues/82

gi0baro:
Close #404

### DIFF
--- a/asgiref/typing.py
+++ b/asgiref/typing.py
@@ -32,6 +32,7 @@ __all__ = (
     "HTTPResponseStartEvent",
     "HTTPResponseBodyEvent",
     "HTTPResponseTrailersEvent",
+    "HTTPResponsePathsendEvent",
     "HTTPServerPushEvent",
     "HTTPDisconnectEvent",
     "WebSocketConnectEvent",
@@ -136,6 +137,11 @@ class HTTPResponseTrailersEvent(TypedDict):
     type: Literal["http.response.trailers"]
     headers: Iterable[Tuple[bytes, bytes]]
     more_trailers: bool
+
+
+class HTTPResponsePathsendEvent(TypedDict):
+    type: Literal["http.response.pathsend"]
+    path: str
 
 
 class HTTPServerPushEvent(TypedDict):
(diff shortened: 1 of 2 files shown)


Kludex:
Not related: From this PR I noticed that Zero Copy Send related type is missing.

gi0baro:
Any chance to get this reviewed?

andrewgodwin (approved):

andrewgodwin:
Sorry - has been a busy few weeks!

PR #409, opened 2023-09-22
Commented scope that caused error with unix sockets

cgunadi2018:


andrewgodwin:
You can't just remove functional code like this without suggesting a replacement - for a valid PR, I need to see something that preserves the existing behaviour _and_ fixes any potential unix socket issues.

cgunadi2018:
Sorry about this.
This is meant to be my own internal experiment.
I know no one would ever approve something like this, I wouldn't.
Please ignore.

andrewgodwin:
No problem - it happens to us all!

PR #411, opened 2023-10-06
Handle `"client": None` in WsgiToAsgi adapter

### LINKED ISSUE #410: asgiref's WsgiToAsgi wrapper does not handle "client": None per spec
The `WsgiToAsgi` adapter in `asgiref/wsgi.py` raises an exception if it gets an HTTP scopes with `"client": None`.

```
from asgiref.wsgi import WsgiToAsgi
```

The relevant section of code from the repository is shown below.

https://github.com/django/asgiref/blob/2ddc82c1bd658f247d19f62724982941d97800f3/asgiref/wsgi.py#L79-L80

Line 80 raises a `KeyError` if "client" in the scope has value None, since it tries to index a NoneType with index `0`.

---

The [latest version of the spec](https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope) describes the behavior of this part of the scope:

> `client` (Iterable[Unicode string, int]) – A two-item iterable of `[host, port]`, where `host` is the remote host’s IPv4 or IPv6 address, and `port` is the remote port as an integer. Optional; if missing defaults to `None`.

As a workaround, servers can just not set `"client"` at all. But it's surprising that WsgiToAsgi doesn't handle the case where `"client"` is set to `None` because that is indicated as the default behavior in the spec.

ekzhang:
This fixes #410. Please see the issue for a description of the change.

### DIFF
--- a/asgiref/wsgi.py
+++ b/asgiref/wsgi.py
@@ -76,7 +76,7 @@ class WsgiToAsgiInstance:
             environ["SERVER_NAME"] = "localhost"
             environ["SERVER_PORT"] = "80"
 
-        if "client" in scope:
+        if scope.get("client") is not None:
             environ["REMOTE_ADDR"] = scope["client"][0]
 
         # Go through headers and make them into environ entries


andrewgodwin (approved):

andrewgodwin:
Thanks for catching this!

PR #418, opened 2023-11-06
Add Connexion to implementations.rst

RobbeSneyders:
[Connexion](https://github.com/spec-first/connexion) has joined the ASGI ecosystem with its 3.0.0 release in which it adopted the ASGI interface.


andrewgodwin:
Amazing - glad to hear it!

PR #432, opened 2024-01-13
Test asyncio.shield with sync and async middlware

ttys0dev:
I noticed asyncio.shield doesn't work as expected when using sync middleware, I've wrote two tests, one which passes for async middleware and one which fails for sync middleware.

### DIFF
--- a/tests/test_sync.py
+++ b/tests/test_sync.py
@@ -849,3 +849,122 @@ async def test_sync_to_async_within_create_task():
         await sync_to_async(sync_middleware)()
 
     assert task_executed
+
+
+@pytest.mark.asyncio
+@pytest.mark.skip(reason="deadlocks")
+async def test_inner_shield_sync_middleware():
+    """
+    Tests that asyncio.shield is capable of preventing http.disconnect from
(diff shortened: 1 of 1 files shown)


andrewgodwin (approved):

andrewgodwin:
The code looks reasonable and I'll always take more tests - could you add a docstring to each test that describes exactly the situation it's testing?

ttys0dev:
> could you add a docstring to each test that describes exactly the situation it's testing?

Added some more details/comments, how does that look?

andrewgodwin:
They're better, but I'm a little bit on the fence about including Django-specific tests in asgiref - I know they're not actually that Django specific as they're testing sync_to_async and async_to_sync, but the way they're phrased is a bit specific - that said, they're still good tests. I think I'll merge them in and then just modify the docstring a little to say "django style" or something.

Thanks for putting these together!

ttys0dev:
> the way they're phrased is a bit specific

I mostly just tried to copy the style of the existing django style tests.

By the way should I open a tracking issue for the disabled deadlocking test?

Issue #434, opened 2024-01-14
Using asyncio.shield hangs/deadlocks when used with sync middleware.


ttys0dev:
There is a deadlock bug which is demonstrated by [this](https://github.com/django/asgiref/blob/19e14e7ab91a5f03b7b32dccae69f74933399cb4/tests/test_sync.py#L854-L912) recently added test(added in #432) which is currently disabled. There was another recently fixed deadlock issue in #348 which I'm thinking may possibly be related/similar although I'm not entirely sure.
```python
@pytest.mark.asyncio
@pytest.mark.skip(reason="deadlocks")
async def test_inner_shield_sync_middleware():
    """
    Tests that asyncio.shield is capable of preventing http.disconnect from
    cancelling a django request task when using sync middleware.

    Currently this tests is skipped as it causes a deadlock.
    """

    # Hypothetical Django scenario - middleware function is sync
    def sync_middleware():
        async_to_sync(async_view)()

    task_complete = False
    task_cancel_caught = False

    # Future that completes when subtask cancellation attempt is caught
    task_blocker = asyncio.Future()

    async def async_view():
        """Async view with a task that is shielded from cancellation."""
        nonlocal task_complete, task_cancel_caught, task_blocker
        task = asyncio.create_task(async_task())
        try:
            await asyncio.shield(task)
        except asyncio.CancelledError:
            task_cancel_caught = True
            task_blocker.set_result(True)
            await task
            task_complete = True

    task_executed = False

    # Future that completes after subtask is created
    task_started_future = asyncio.Future()

    async def async_task():
        """Async subtask that should not be canceled when parent is canceled."""
        nonlocal task_started_future, task_executed, task_blocker
        task_started_future.set_result(True)
        await task_blocker
        task_executed = True

    task_cancel_propagated = False

    async with ThreadSensitiveContext():
        task = asyncio.create_task(sync_to_async(sync_middleware)())
        await task_started_future
        task.cancel()
        try:
            await task
        except asyncio.CancelledError:
            task_cancel_propagated = True
        assert not task_cancel_propagated
        assert task_cancel_caught
        assert task_complete

    assert task_executed
```

When I run this broken test in a debugger the `asyncio.CancelledError` is incorrectly caught by [this exception block](https://github.com/django/asgiref/blob/19e14e7ab91a5f03b7b32dccae69f74933399cb4/tests/test_sync.py#L906-L907) and the test then hangs/deadlocks immediately after.

[...]

PR #435, opened 2024-01-15
Fix task cancellation propagation to subtasks when using sync middleware

ttys0dev:
Potential fix for #434.


andrewgodwin:
Looks pretty good to me - I'll see how it meshes with #439 as well to make sure that everything is reasonable before I cut a release.

PR #439, opened 2024-01-19
Move variable initialization in AsyncToSync from __init__ to __call__

### LINKED ISSUE #438: async_to_sync does not use the correct thread when used as decorator
When using a function decorated with `@async_to_sync` inside of a function decorated with `@sync_to_async` a new thread is spawned instead of reusing the same thread and event loop. This issue does not happen when doing `coroutine = async_to_sync(coroutine)`.

This is due to the fact that the AsyncToSync sets `main_event_loop` and `main_event_loop_pid` on `__init__` and not on `__call__` and SyncToAsync only sets  `main_event_loop` and `main_event_loop_pid` on `__call__`.

The easy way to solve this is moving the initialization of `self.main_event_loop` and `self.main_event_loop_pid` from `__init__` to `__call__`  in AsyncToSync (I can make a PR if needed)

Consider this example on why this is relevant:
```python
import asyncio
from asgiref.sync import sync_to_async, async_to_sync
import threading

# this function cannot be changed
def lib_call(callback):
    callback()

def callback():
    async def coro():
        print(f"Callback thread: {threading.get_ident()}")
        # async code here
    coro_sync = async_to_sync(coro)
    coro_sync()
 
@async_to_sync   
async def callback_decorated():
    print(f"Callback decorated thread: {threading.get_ident()}")
    # async code here
    
@sync_to_async
def blocking_func():
    print(f"Blocking func thread: {threading.get_ident()}")
    # this spawns a new thread and new event loop for the callback
    lib_call(callback)
    # this uses the outer event loop
    lib_call(callback_decorated)

async def main():
    try:
        async with asyncio.TaskGroup() as tg:
            task0 = tg.create_task(blocking_func())
    except Exception as exc:
        pass
    
print(f"Main thread: {threading.get_ident()}")
asyncio.run(main())
```

The output is:
```
Main thread: 140545354248192
Blocking func thread: 140545298830912
Callback thread: 140545354248192
Callback decorated thread: 140544949696064
```

germaniuss:
Closes #438


andrewgodwin:
OK, so the code still looks good, but since this code is best tested to within an inch of its life - think you can write a test that ensures this behaviour is there in future?

germaniuss:
> OK, so the code still looks good, but since this code is best tested to within an inch of its life - think you can write a test that ensures this behaviour is there in future?

Sure thing

germaniuss:
Done!

andrewgodwin:
I think the test naming is fine - just satisfy the linter, and I'll land it!

ttys0dev:
> Wait, 3.8 failed, womp womp.

Well we can just disable that specific test that's breaking on 3.8 like in #440.

andrewgodwin:
Merged in the form of #440 instead.

Issue #459, opened 2024-04-08
Spec question: `websocket.disconnect` doesn't support the `reason` field


frankie567:
From the WebSocket specification, it's possible to set a "reason" when either the server or the client closes the connection: https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.6

In the ASGI specification, this is well supported for the "Close - `send` event" (i.e. close sent by the server): 

https://github.com/django/asgiref/blob/db3ff43e9fa1daf7c6b1cb67db8eec1885be911d/specs/www.rst?plain=1#L476-L496

However, for the "Disconnect - `receive` event (i.e. close sent by the client), it is not supported; only the code can be set:

https://github.com/django/asgiref/blob/db3ff43e9fa1daf7c6b1cb67db8eec1885be911d/specs/www.rst?plain=1#L440-L458

Is there a particular reason to omit this field for this particular event?

andrewgodwin:
I do not remember a specific reason; it was likely either an oversight or the libraries I developed the spec against last decade didn't support it.

frankie567:
Thank you for your feedback 🙂 Then, are you okay if I propose a PR to add this to the spec?

Kludex:
Can you give me some days to reply here? I'm on vacation, but I'm back next week.

frankie567:
Sure! Enjoy your holidays 🌴

frankie567:
Hey @Kludex 👋  Did you have some time to think about this? 🙂

Kludex:
It looks like an oversight, but... No one complained in years... Do you need it?

frankie567:
Well, not a *really* big issue but yes, I had one complain downstream in `httpx-ws`. Relevant context:

* https://github.com/encode/uvicorn/discussions/2299
* https://github.com/frankie567/httpx-ws/issues/70

IMHO, it looks like a harmless addition that would make the spec a bit more compliant regarding WS spec. Happy to make the PR if. you agree 🙂

Kludex:
Are the WebSockets clients sending the "reason"?

frankie567:
They can, if I understand correctly the [spec](https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.6). It's apparently supported by browsers API: https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/close

Kludex:
Then I don't mind adding it.

PR #462, opened 2024-05-06
Spec: Added `reason` to WebSocket Disconnect

frankie567:
This is a proposal to fix the discussion in #459

Clients can send this reason field when disconnecting. This is allowed as per RFC 6455 and supported in browsers JS API, but previously not supported in ASGI.

---

I took the liberty to bump the spec version, but I can remove it if you prefer 🙂

### DIFF
--- a/asgiref/typing.py
+++ b/asgiref/typing.py
@@ -189,6 +189,7 @@ class WebSocketResponseBodyEvent(TypedDict):
 class WebSocketDisconnectEvent(TypedDict):
     type: Literal["websocket.disconnect"]
     code: int
+    reason: Optional[str]
 
 
 class WebSocketCloseEvent(TypedDict):
--- a/specs/www.rst
+++ b/specs/www.rst
@@ -2,7 +2,7 @@
 HTTP & WebSocket ASGI Message Format
 ====================================
 
-**Version**: 2.4 (2024-01-16)
+**Version**: 2.5 (2024-06-05)
 
 The HTTP+WebSocket ASGI sub-specification outlines how to transport HTTP/1.1,
 HTTP/2 and WebSocket connections within ASGI.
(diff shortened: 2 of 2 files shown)


asgiref/typing.py
```diff
@@ -189,6 +189,7 @@ class WebSocketResponseBodyEvent(TypedDict):
 class WebSocketDisconnectEvent(TypedDict):
     type: Literal["websocket.disconnect"]
     code: int
+    reason: Optional[str]
```
Kludex, line 192: @andrewgodwin Can we start using `NotRequired` on those?

specs/www.rst
```diff
@@ -22,6 +22,7 @@ This spec has had the following versions:
 * ``2.2``: Allow ``None`` in the second item of ``server`` scope value.
 * ``2.3``: Added the ``reason`` key to the WebSocket close event.
 * ``2.4``: Calling ``send()`` on a closed connection should raise an error
+* ``2.5``: Added the ``reason`` key to the WebSocket disconnect event.
```
Kludex, line 25: Does this need a bump on the spec?

frankie567: Not for me to say, but I saw it was done similarly for 2.3. Can remove it if not needed :)

frankie567:
Anything I can do so we can go forward with this? 🙂

Kludex (approved):

asgiref/typing.py
andrewgodwin: Unfortunately that only appears in 3.11...

specs/www.rst
andrewgodwin: Yes, any spec change should minor bump if it's backwards-compatible (like this) and a major bump if it is not.

andrewgodwin (approved):

specs/www.rst
andrewgodwin: You did miss out updating the main version number though - we'll need that bumped too.

andrewgodwin (changes requested):

specs/www.rst
frankie567: > You did miss out updating the main version number though - we'll need that bumped too.

Done 👍

andrewgodwin (approved):

andrewgodwin:
Thanks for your help on this!

Kludex:
@andrewgodwin can you push an update to the docs, please?

andrewgodwin:
It was meant to auto-build but it seems the webhook became unhappy - just initiated a manual build, so it should be good soon.

Kludex:
> It was meant to auto-build but it seems the webhook became unhappy - just initiated a manual build, so it should be good soon.

Thanks!

PR #465, opened 2024-08-08
Update HTTP spec: correct `scope.asgi.spec_version`

lazorchakp:
I noticed a tiny gap in the ASGI HTTP spec for `scope["asgi"]["spec_version"]` - all valid versions are explicitly enumerated, but that list hasn't been updated to include newer versions of the ASGI HTTP spec.

This change rewords the spec so the list of versions doesn't need to be updated every time there's a new version of the ASGI HTTP spec. I'd also be happy to just add the newer versions to this list instead, but then someone would also need to remember to update it in the future too.

I tested this by rebuilding the docs locally and viewing the changes in my browser:

image

### DIFF
--- a/specs/www.rst
+++ b/specs/www.rst
@@ -90,8 +90,8 @@ The *connection scope* information passed in ``scope`` contains:
 * ``asgi["version"]`` (*Unicode string*) -- Version of the ASGI spec.
 
 * ``asgi["spec_version"]`` (*Unicode string*) -- Version of the ASGI
-  HTTP spec this server understands; one of ``"2.0"``, ``"2.1"``, ``"2.2"`` or
-  ``"2.3"``. Optional; if missing assume ``2.0``.
+  HTTP spec this server understands; for example: ``"2.0"``, ``"2.1"``, ``"2.2"``,
+  etc. Optional; if missing assume ``"2.0"``.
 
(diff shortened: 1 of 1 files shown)


specs/www.rst
```diff
@@ -90,8 +90,8 @@ The *connection scope* information passed in ``scope`` contains:
 * ``asgi["version"]`` (*Unicode string*) -- Version of the ASGI spec.
 
 * ``asgi["spec_version"]`` (*Unicode string*) -- Version of the ASGI
-  HTTP spec this server understands; one of ``"2.0"``, ``"2.1"``, ``"2.2"`` or
-  ``"2.3"``. Optional; if missing assume ``2.0``.
+  HTTP spec this server understands; for example: ``"2.0"``, ``"2.1"``, ``"2.2"``,
+  etc. Optional; if missing assume ``"2.0"``.
```
Kludex, line 94: Why the etc? It's just those 4 values 👀 
```suggestion
 ``"2.3"``. Optional; if missing assume ``"2.0"``.
```

lazorchakp: Can't it also be `2.4` or `2.5` now? It's entirely possible I've misread the spec. if so I'll close this; I agree that if those are the only valid values, this change is unnecessary.

andrewgodwin: Yes, the www spec is now up to 2.5, so I think this makes sense.

PR #471, opened 2024-09-29
Fix #470: Prevent error `AsyncToSync.main_wrap() got multiple values for argument '<kwarg>'`

### LINKED ISSUE #470: `AsyncToSync` incorrectly passes arguments when called function arguments to internal wrapper
If the function being wrapped uses a kwarg that has the same name as the `AsyncToSync.main_wrap` function, it causes the execution to fail with `AsyncToSync.main_wrap() got multiple values for argument '<kwarg>'`.

How to reproduce - run the following example:

```python
from asgiref.sync import async_to_sync

@async_to_sync
async def example(context):
    pass

example(context=1)
```

Stacktrace:
```
  File "example.py", line 9, in <module>
    example(context=1)
  File "<venv>/asgiref/sync.py", line 215, in __call__
    awaitable = self.main_wrap(
                ^^^^^^^^^^^^^^^
TypeError: AsyncToSync.main_wrap() got multiple values for argument 'context'
```

I have tested this on latest `main`, on 3.8.1, 3.7.2, 3.7.1, and 3.7.0 versions of asgiref.
From my investigation, the culprit is #390. I tested on the commit before that PR (fd7a63c591653146379b55e68a29883b9bf7e706) and the bug does not reproduce. So the last working version which does not exhibit the bug is 3.6.0.

Krismix1:
Fixes #470.

Destructuring `*args` and `**kwargs` passed in by the callee has the chance that the same kwargs names may be used as by `main_wrap`. If that happens, the interpreter will raise a error like: `TypeError: AsyncToSync.main_wrap() got multiple values for argument 'context'`.
`ParamSpec` requires the `*args` and `**kwargs` syntax to be used, but it is wrong execution wise.
So we work around the issue by pre-creating the awaitable outside `main_wrap`, which removes the need to pass `*args` and `**kwargs`.
Another alternative instead of creating the awaitable could be to use `functools.partial`, though that does add some performance overhead: https://github.com/django/asgiref/issues/470#issuecomment-2377074109.

### DIFF
--- a/asgiref/sync.py
+++ b/asgiref/sync.py
@@ -212,17 +212,17 @@ class AsyncToSync(Generic[_P, _R]):
         # main event loop's thread if it's there, otherwise make a new loop
         # in this thread.
         try:
-            # make a function taking no args/kwargs which can be passed as is
-            # to self.main_wrap, so that `args` and `kwargs` don't need to be
-            # destructured when passed to self.main_wrap
-            # (which is required by `ParamSpec`)
-            awaitable_fn = functools.partial(self.awaitable, *args, **kwargs)
             awaitable = self.main_wrap(
                 call_result,
                 sys.exc_info(),
                 task_context,
                 context,
-                awaitable_fn,
+                # prepare an awaitable which can be passed as is to self.main_wrap,
+                # so that `args` and `kwargs` don't need to be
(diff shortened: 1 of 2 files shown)


andrewgodwin:
That's a very neat way of solving it - nice! And thanks!

Issue #468, opened 2024-09-17
Clarify `raw_path` Definition for WebSockets Protocol in the Specification


raptium:
The definition of `raw_path` in the ASGI specification differs slightly between HTTP and WebSockets protocols.

- **HTTP**: The specification explicitly states that the query string is excluded from `raw_path`.

https://github.com/django/asgiref/blob/b7aaa795fbbea7ca88458279cf8302533d249101/specs/www.rst?#L107-L110

- **WebSockets**: The specification does not mention whether the query string should be included or excluded from `raw_path`.

https://github.com/django/asgiref/blob/b7aaa795fbbea7ca88458279cf8302533d249101/specs/www.rst?#L318-L321

To ensure clarity and consistency, it would be beneficial to explicitly define the behavior of `raw_path` for WebSockets:

1. **If the query string should be excluded**: The specification should use the same wording as in the HTTP section to avoid ambiguity.
2. **If the query string should be included**: This should be explicitly stated, along with any rationale for the differing behavior.

Given the potential for confusion, it would be helpful to clarify this in the specification.

Kludex:
Yeah, having this difference looks a bit weird. Do you remember if it was intentional? @andrewgodwin

andrewgodwin:
I do not believe that was intentional - my guess is that we specifically clarified it for standard HTTP but not for WebSocket. I would say we should update the WebSocket side to say that it also excludes the query string.

Kludex:
@raptium Would you like to send a PR here?

Kludex:
We can close this @andrewgodwin 🙏

PR #474, opened 2024-10-04
Clarify `raw_path` of WebSocket also excludes query string

raptium:
Discussed in #468, `raw_path` of WebSocket is updated for clarification.


Kludex (approved):

andrewgodwin:
Thanks for this!

PR #481, opened 2024-10-09
Simplify local's implementation

spanezz:
Note: this PR applies over #478.

This is a proposal for a simplification of `asgiref.local.Local` as a thin adapter layer over either `threading.local` or `ContextVar`.

### DIFF
--- a/asgiref/local.py
+++ b/asgiref/local.py
@@ -1,5 +1,3 @@
-import asyncio
-import contextlib
 import contextvars
 import threading
 from typing import Any, Union
@@ -9,13 +7,15 @@ class _CVar:
     """Storage utility for Local."""
 
     def __init__(self) -> None:
+        self._thread_lock = threading.RLock()
         self._data: dict[str, contextvars.ContextVar[Any]] = {}
 
     def __getattr__(self, key: str) -> Any:
-        try:
-            var = self._data[key]
-        except KeyError:
-            raise AttributeError(f"{self!r} object has no attribute {key!r}")
+        with self._thread_lock:
+            try:
+                var = self._data[key]
+            except KeyError:
+                raise AttributeError(f"{self!r} object has no attribute {key!r}")
 
         try:
             return var.get()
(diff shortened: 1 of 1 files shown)


andrewgodwin:
I get the idea of this, and I think it makes sense, but this is deep in the part of the codebase that we can land a big change in, all be confident that it works great, and then it mysteriously breaks Django right afterwards.

Are there any downsides in your mind to redoing it like this?

spanezz:
> I get the idea of this, and I think it makes sense, but this is deep in the part of the codebase that we can land a big change in, all be confident that it works great, and then it mysteriously breaks Django right afterwards.
> 
> Are there any downsides in your mind to redoing it like this?

The only downside that I can think of is that I might be missing something important, and my approach is wrong. I tried to follow the effect of the code also on Django's main branch, and I still couldn't see downsides.

Especially when some infrastructure is deep and core, I feel motivated to try and make it as minimal and clearly defined as possible. I find it tends to pay off in the medium-long term, as each extra code or layer makes things harder to reason and debug, and it tends to be more critical the more core a piece of code is. When studying the code for #478 I spotted this opportunity and tried to explore it.

I'd say that the code simplification, if it turns out to be valid, would ripple in the definition of `thread_critical` as an argument to `Local`: if `thread_critical=True`, the contents of the Local are shared between asyncio tasks, and if `thread_critical=False` asyncio tasks inherit the Local values but the changes they perform on it are isolated.

The asgiref test suite would agree with this, and `thread_critical` is not used by other asgiref code. Django uses it only in `django.db.utils.ConnectionHandler` to, as I understand it, prevent reusing the same database handle between different threads, while cross-talk between different asyncio tasks is ok. Indeed if a Task opens a DB connection, it's ok if other tasks reuse it.

If I read that correctly, then Django's `ConnectionHandler` could be refactored to use a `threading.local`, reducing the abstraction load needed to follow that critical code. `asgiref.local.Local` could then eventually loose `thread_critical`, and focus on being a `threading.local` extension that also isolates asyncio Tasks.

I would understand if this is a can of worms you don't want to open, though to me they look like yummy worms :)

PR #497, opened 2025-02-28
False deadlock detection in case of exception in run_in_executor

namezys:
See issue #495 

I suppose this problem can only affect developing/debugging because it should not happen in production.
In my case, the executor was shut down unexpectedly, but I spent time to find the deadlock.
Also, a real exception will be raised only one time, and you will get deadlock detection further, which is more confusing.

However, I have doubts about catching a cancelation error here like in awaiting. A quick review of PEPs didn't answer.


carltongibson:
Closing as per #495

PR #507, opened 2025-04-11
Avoid using event_loop.is_running()

graingert:
* `asyncio.get_running_loop().is_running()` is redundant
* checking `is_running` before `call_soon_threadsafe` is redundant because `call_soon_threadsafe` checks it, and the loop could have stopped between the check and the call


carltongibson (approved):
Yes, nice. 

As well, I like that this makes the boolean logic main event loop first. The old `if not ( ... and ... ): ... else: ...` is confusing (at least to me).

PR #511, opened 2025-04-26
added AsyncSingleThreadContext

Arfey:
I have a problem with async database connections for `Django` `TestCase`.

The psycopg3 connection is not thread-safe. As a result, I need to create a new connection per thread. This is a problem for the `TestCase` coz it uses a transaction for each test case.

```python
class TestCase(TransactionTestCase):
    async def _apost_teardown(self):
        for conn in async_connections.all(initialized_only=True):
            await conn.close()

    def _setup_and_call(self, result, debug=False):
        if iscoroutinefunction(testMethod):
            ############### convert test 
            setattr(self, self._testMethodName, async_to_sync(testMethod))

        if debug:
            super().debug()
        else:
            super().__call__(result)

       self._post_teardown()
       ############### convert teardown 
       async_to_sync(self._apost_teardown)()       
```

I can do something like that coz each async_to_sync function uses a new thread executor (and different connection as a result). To solve that, I try to implement something similar for ThreadSensitiveContext.

```python
class TestCase(TransactionTestCase):
    def _setup_and_call(self, result, debug=False):
        with AsyncSingleThreadContext()
            if iscoroutinefunction(testMethod):
                ############### convert test 
                setattr(self, self._testMethodName, async_to_sync(testMethod))

            if debug:
                super().debug()
            else:
                super().__call__(result)

           self._post_teardown()
           ############### convert teardown 
           async_to_sync(self._apost_teardown)()       
```

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,10 @@
+3.10.0 (UNRELEASED)
+-------------------
+
+* Added AsyncSingleThreadContext context manager to ensure multiple AsyncToSync
+  invocations use the same thread. (#511)
+
+
 3.9.2 (2025-09-23)
 ------------------
 
--- a/asgiref/sync.py
+++ b/asgiref/sync.py
@@ -69,6 +69,45 @@ else:
         return func
 
 
+class AsyncSingleThreadContext:
+    """Context manager to run async code inside the same thread.
+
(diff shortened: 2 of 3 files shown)


fcurella:
Hi! how can we help getting this reviewed?

PR #516, opened 2025-06-29
chore: remove not needed anymore typing workaround

bellini666:
I introduced this workaround in #390 due to the reasons stated in the comment.

As the minimum Python version for this lib is 3.9 now, the workaround can be safely removed


carltongibson (approved):
Thanks for this @bellini666 👍

PR #530, opened 2025-09-15
Fixed #529: wsgi.errors should be a StringIO

### LINKED ISSUE #529: wsgi.errors should be a text stream, not a binary stream
Causing issues with Flask's logging handler (it writes to `wsgi.errors`)

```python
--- Logging error ---
  File "/home/arthur/.local/share/uv/python/cpython-3.13.0-linux-x86_64-gnu/lib/python3.13/logging/__init__.py", line 1153, in emit
    stream.write(msg + self.terminator)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
TypeError: a bytes-like object is required, not 'str'
Call stack:
[...]
  File "/home/arthur/Documents/work/myapp/myapp/app_logging.py", line 37, in _log_request_result
    logger.info("Request complete", extra=data)
Message: 'Request complete'
Arguments: ()
```

Offending line:
https://github.com/django/asgiref/blob/fbe9506df140a437921a01ea824dbe51d8c38c24/asgiref/wsgi.py#L70

[WSGI spec](https://peps.python.org/pep-3333/#environ-variables):

> An output stream (file-like object) to which error output can be written, [...]. This should be a “text mode” stream; i.e., applications should use "\n" as a line ending [...]
> (On platforms where the str type is unicode, the error stream should accept and log arbitrary unicode without raising an error; it is allowed, however, to substitute characters that cannot be rendered in the stream’s encoding.)

---

Why wasn't this raised sooner ?
Flask has been using `wsgi.errors` for >8 years. This repo has been using BytesIO since WsgiToAsgi exists (8 years).
 
---

Workaround:

```py
import sys
import io

from asgiref.wsgi import WsgiToAsgi, WsgiToAsgiInstance


class WsgiToAsgiInstanceFixed(WsgiToAsgiInstance):
    def build_environ(self, scope, body):
        environ = super().build_environ(scope, body)
        # environ["wsgi.errors"] = io.StringIO() # Doesn't print anything ?
        environ["wsgi.errors"] = sys.stderr
        return environ

class WsgiToAsgiFixed(WsgiToAsgi):
    async def __call__(self, scope, receive, send):
        await WsgiToAsgiInstanceFixed(self.wsgi_application)(scope, receive, send)
```

andrewgodwin:
This fixes the direct incompatibility that `wsgi.errors` should be a string-based file object rather than the bytes-based one it has been for 8 years - Flask/werkzeug expects to write to this.

This points it at `sys.stderr` as that's probably the best option ASGI has for this info.

### DIFF
--- a/asgiref/wsgi.py
+++ b/asgiref/wsgi.py
@@ -1,4 +1,4 @@
-from io import StringIO
+import sys
 from tempfile import SpooledTemporaryFile
 
 from asgiref.sync import AsyncToSync, sync_to_async
@@ -67,7 +67,7 @@ class WsgiToAsgiInstance:
             "wsgi.version": (1, 0),
             "wsgi.url_scheme": scope.get("scheme", "http"),
             "wsgi.input": body,
-            "wsgi.errors": StringIO(),
+            "wsgi.errors": sys.stderr,
             "wsgi.multithread": True,
             "wsgi.multiprocess": True,
             "wsgi.run_once": False,


awoimbee (approved):
Thanks, I didn't have the courage to change such old code !
gunicorn writes to `sys.stderr` too (+ `gunicorn.error` handlers) so this seems to be correct !

https://github.com/benoitc/gunicorn/blob/a86ea1e4e6c271d1cd1823c7e14490123f9238fe/gunicorn/http/wsgi.py#L71

PR #531, opened 2025-09-17
Fix incorrect Trailer header

anuraaga:
The header specifying which trailers will be sent is `Trailer`, not `Trailers`

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Trailer


andrewgodwin:
Thanks!

PR #537, opened 2025-11-26
Fix pass callable object to sync_to_async

Maillol:
### Fix: Prevent passing callable objects that overwrite context

This fixes a regression where passing a callable object like huey.Task would break the context attribute.

After this change, huey.Task and similar callables can be used without overwriting context, restoring previous behavior.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,9 @@
+UNRELEASED
+----------
+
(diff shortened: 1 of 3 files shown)


carltongibson:
Good example @Maillol. This looks right. I'll try and pull a point release together shortly.

carltongibson (approved):
Thanks @Maillol. I rebased and added a release note. Planning a release next week.

Issue #423, opened 2023-11-29
Re-evaluate "http.response.zerocopysend" 


Archmonger:
Right now, none of the existing ASGI webservers support `zerocopysend`. This [seems to be](https://github.com/encode/uvicorn/pull/1210#issuecomment-986651500) due to the additional code complexity needed to support this extension.

This lack of support has prevented me from adding `zerocopysend` to [whitenoise](https://github.com/evansd/whitenoise/pull/359#discussion_r801136794), which would have been a big win for the Python web development community.

I'm opening this issue as a place to discuss what, if anything, can be done to encourage adoption for `zerocopysend`.

Here's a few questions to get this discussion started:

1. Is a `zerocopysend2` interface needed to encourage adoption?
2. Does `asgiref` need some helper utilities to simplify adoption of the current extension?
3. It seems that adding `zerocopysend` would require [facing the buffer](https://github.com/encode/uvicorn/pull/1210#issuecomment-986705426). Is it worthwhile (or feasible) to engineer a way of avoiding this?

Archmonger:
cc: @pgjones @Kludex @tomchristie

Kludex:
The ones who know more about it are @synodriver and @abersheeran.

abersheeran:
The reason for facing buffers is because of the design of asyncio. It seems possible to solve this problem if we modify the zerocopysend standard so that it can only be sent at the end of the message.

Archmonger:
@abersheeran

In terms of what this would like in practice, do you have a proposed interface?

Trying to work backwards and see how/if to push through this issue.

synodriver:
I have a fork of [hypercorn](https://github.com/nonebot/nonecorn) with this extension, but it will only work under http1.1, since [h2](https://github.com/python-hyper/h2/issues/236) doesn't provide this feature. As for facing the buffer, I personally think there's no problem because mine has been running for a long time.

Archmonger:
@synodriver 

Would an interface change allow h2 support without waiting on library support?

synodriver:
> @synodriver
> 
> Would an interface change allow h2 support without waiting on library support?

I don't think it's possible. hypercorn relys on those sans-io projects to parse raw data. You can send data directly, but that will break h2 library's internal state machine.

Archmonger:
My immediate family impression it's debatable whether http2 and http3 file transfers should be funneled through external libraries.

It seems fairly straightforward and possible to handle them outside of external library support.

[...]

PR #538, opened 2025-12-06
Added offset and count parameters to pathsend extension

### LINKED ISSUE #469: Add `offset` and `count` to Path Send
This discussion exists to talk about whether the Path Send spec can be reasonably modified to support `offset` and `count`, in a similar manner to those respective variables within `zerocopysend`.

As a summary of the [original discussion](https://github.com/django/asgiref/issues/423#issuecomment-2365099879)...

- `zerocopysend` was effectively dead on arrival due to integration challenges with existing webservers.
- `pathsend` can be considered as an alternative that is more compatible with a variety of webservers.
- `pathsend` currently does not support `offset` and `count`, which makes it impossible to use with HTTP range requests.
- `andrewgodwin` agreed that `offset` and `count` can be added to `pathsend` if it can be implemented in a backwards compatible manner
- I suggested if backwards compatibility is too challenging, then the extension namespace can be cluttered with `http.response.pathsend2`

matthewgrossman:
### Background Context

I read #469 and discussed with @gi0baro in this thread https://github.com/emmett-framework/granian/discussions/157 about trying to get this as part of the ASGI spec. The implementation here is just a straight copy of the discussion in the relevant thread. It didn't seem that people had too much of a disagreement of what it'd look like, and if we did we could hash it out in this PR.

Links:
- https://github.com/django/asgiref/issues/423: Issue describing problems with the current zerocopysend
- #469: Issue describing potential improvements to pathsend (what this PR implements)
- https://github.com/emmett-framework/granian/discussions/157: discussion that led me to making this PR in this repo

---

(draft)

This extends the `http.response.pathsend` extension to support partial file sends by adding optional offset and count parameters, enabling HTTP Range request support.

The parameters mirror the existing `http.response.zerocopysend` API:
- `offset`: byte position to start reading from (default: beginning)
- `count`: number of bytes to send (default: to end of file)

This allows ASGI frameworks to implement range requests without loading the bytes themselves, offloading the operation to the server.

Servers may advertise range support capability by setting {"ranges": True} in the extension scope dict, allowing frameworks to detect this capability.

The application remains responsible for setting appropriate headers (Content-Range, Content-Length, Accept-Ranges) in the response.

Fixes #469

---

### Implementation

This original [comment](https://github.com/django/asgiref/issues/469#issuecomment-2373742600) also proposed an alternative that uses `range` instead of offset+count. I don't really have strong opinions one way or another about which we use, but I see the trade-offs as:

1. `range` is just one key, and its explicit about the start/end bytes. If we were to do this, we'd need to make a decision about being inclusive/exclusive with the ending byte (I think exclusive ending byte is common in most situations, but http range GETs are inclusive, so we should likely match that: [docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Range_requests#requesting_a_specific_range_from_a_server))
2. offset/count matches what we have for zerocopysend, so we're staying consistent within ASGI

I'm also in the process of a branch on granian where I'm going to do a quick example implementation just to give an example. I'll likely use granian + starlette to showcase what this extended pathsend could look like

First time contributing to and submitting a PR to asgiref, so please feel free to badger me about bits that I'm missing!

### DIFF
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -146,6 +146,22 @@ ASGI servers that implement this extension will provide
         },
     }
 
+Servers that support the ``offset`` and ``count`` parameters for range
+requests must advertise this capability via::
+
+    "scope": {
+        ...
+        "extensions": {
+            "http.response.pathsend": {
+                "ranges": True,
+            },
+        },
+    }
+
+Applications should check for the presence of ``ranges`` before using
+``offset`` and ``count`` parameters. Servers that advertise
+``"ranges": True`` must handle these parameters correctly.
+
 The ASGI framework can initiate a path-send by sending a message with
 the following keys. This message can be sent at any time after the
 *Response Start* message, and cannot be mixed with ``http.response.body``.
@@ -159,6 +175,14 @@ Keys:
 
 * ``path`` (*Unicode string*): The string representation of the absolute
   file path to be sent by the server, platform specific.
+
+* ``offset`` (*int*): Optional. If this value exists, it will specify
+  the offset at which the server starts to read data from ``path``.
+  Otherwise, it will be read from the beginning of the file.
+
+* ``count`` (*int*): Optional. ``count`` is the number of bytes to
+  copy from the file. If omitted, the file will be read from the
+  offset (or beginning) until its end.
 
 The ASGI application itself is responsible to send the relevant headers
 in the *Response Start* message, like the ``Content-Type`` and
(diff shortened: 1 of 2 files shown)


docs/extensions.rst
```diff
@@ -146,6 +146,22 @@ ASGI servers that implement this extension will provide
...
+        "extensions": {
+            "http.response.pathsend": {
+                "ranges": True,
+            },
+        },
+    }
+
+However, capability signaling is optional; applications may attempt to use
+``offset`` and ``count`` parameters regardless of this flag, and servers
+should handle these parameters if provided.
```
matthewgrossman: wasn't sure on the convention here; if a framework doesn't explicitly convey support for `ranges`, should the server still attempt to read the `offset`/`count` and utilize it if it exists? Or do we generally try to avoid that "fuzzy capability matching" in asgi servers/frameworks?

carltongibson:
Cc @Kludex

Issue #428, opened 2023-12-23
Add Type Annotations to `asgiref` Module Functions


davitacols:
I suggest adding type annotations to the following functions in the `asgiref` module:

1. `is_double_callable`
2. `double_to_single_callable`
3. `guarantee_single_callable`

**Proposed Changes:**
- Add type annotations to function parameters and return types in the specified functions.

**Benefits:**
- Improved code readability.
- Better tooling support for developers using the `asgiref` module.

**Additional Context:**
This change aims to enhance the maintainability and clarity of the `asgiref` codebase by incorporating type annotations into key functions.

andrewgodwin:
Sounds reasonable - I'll happily take a pull request for this.

davitacols:
I am happy you found this useful.
Best.

Joao-antonio-gg:
Hello, can i try this Pr?

davitacols:
> Hello, can i try this Pr?

Pls go  ahead

patx:
https://github.com/django/asgiref/pull/541

PR #541, opened 2025-12-18
Add type annotations to asgiref module. 

patx:
https://github.com/django/asgiref/issues/428

### DIFF
--- a/asgiref/compatibility.py
+++ b/asgiref/compatibility.py
@@ -1,7 +1,8 @@
 from __future__ import annotations
 
 import inspect
-from typing import Any, Awaitable, Callable, Union
+from typing import Any, Awaitable, Callable, Union, cast
+
 from typing_extensions import TypeAlias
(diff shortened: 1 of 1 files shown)


davitacols:
I tried a follow-up locally on top of this PR.

One improvement is to reuse the existing aliases from `asgiref.typing` instead of defining local aliases in `compatibility.py`, and to type `is_double_callable()` as `TypeGuard[ASGI2Application]` so `guarantee_single_callable()` narrows cleanly.

I pushed that version to my fork here:
https://github.com/davitacols/asgiref/tree/pr-541-typeguard

Commit: `c482872` (`Reuse shared ASGI types in compatibility helpers`)

It passed locally with:
- `python -m pytest -v`
- `python -m mypy asgiref tests setup.py docs/conf.py`

If useful, you can cherry-pick `c482872` or diff it against this PR branch.

PR #542, opened 2026-01-21
Implemented sharing the same event loop with AsyncSingleThreadContext

Arfey:
The pull request updates how the `AsyncSingleThreadContext` works in `asgiref.sync` so that multiple `async_to_sync` calls within the same async context share the same asyncio event loop and thread. This improves consistency and correctness when running async code converted to sync repeatedly in one context.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,6 +36,7 @@ jobs:
           python -m pip install --upgrade tox tox-py
 
       - name: Run tox targets for ${{ matrix.python-version }}
+        timeout-minutes: 10
         run: tox --py current
 
   lint:
--- a/asgiref/sync.py
+++ b/asgiref/sync.py
@@ -8,6 +8,7 @@ import sys
 import threading
 import warnings
 import weakref
+from asyncio import constants
 from concurrent.futures import Future, ThreadPoolExecutor
 from typing import (
     TYPE_CHECKING,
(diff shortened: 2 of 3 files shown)


carltongibson (commented):
Hey @Arfey. Thanks for your patience here, I've been looking at other issues. 

I don't really have any problems with this: you're the only consumer at this point. 

My main concern was churn. We only just added the class, at your request, and then we're immediately changing it. I wonder if it isn't better developed alongside the client code until we're sure it's stabilised, and then we can adopt (and advertise) the working version here? (Not least because otherwise you're waiting on a release here, which I don't imagine doing much more frequently, so end up using a local version anyway 🫠)

Arfey:
Hey 👋,

I think if the public API doesn't change (and it doesn't), then it's not a problem for the end user. The current change just improves the class to allow using a long-running, event-loop-bound client within `AsyncSingleThreadContext`.

I'm not waiting on a release, but I also don't think holding off for years to see how it works before opening a PR is a good idea 😊

carltongibson:
Sure... my question was "is this it, or are there going to be more changes" -- if the latter then I do tend to lean toward working out what's needed first. 

I'll merge this one for the upcoming release.

Arfey:
Oh, got you 😌

To be honest, I've given up on supporting WSGI/ASGI compatibility (that's why we needed this), so I'm not sure whether more changes will follow, since I don't use it anymore.

carltongibson:
Perfect... 🫠

PR #550, opened 2026-02-22
Document Gunicorn as a Python HTTP server

benoitc:
Added Gunicorn section with description and features.


carltongibson:
https://asgi.readthedocs.io/en/latest/implementations.html#gunicorn

PR #558, opened 2026-03-25
Handle stopped main event loops in async_to_sync

### LINKED ISSUE #545: Nested AsyncToSync and SyncToAsync results in blocking code
Hello, 

after updating `asgiref` from version 3.8.1 to version 3.9.0 we've experienced hanging pytest invocations. I suspected a deadlock at first since our code relies on nested `async_to_sync` and `sync_to_async` calls. After running a `git bisect` on `asgiref` I found out that commit https://github.com/django/asgiref/commit/327821b533287d09bd2669c73c1c3d5119ce8bd5 broke our test suite. Particularly, it seems that the removal of the `asyncio.get_running_loop().is_running()` check is responsible for our stalling tests (which the [MR](https://github.com/django/asgiref/pull/507) claims to be redundant). 

The following script will reproduce the error:

```
from asgiref.sync import sync_to_async, async_to_sync

def foo():
    return "foo"

async def foo_async():
    return "foo"

async def async_wrapper():
    return await sync_to_async(foo)()

def test_a_sync_first():
    """
    First sync test - this one passes.
    """
    result = async_to_sync(async_wrapper)()
    assert result == "foo"

def test_b_sync_second():
    """
    Second sync test - this one HANGS!
    """
    result = async_to_sync(foo_async)()
    assert result == "foo"

```

davitacols:
## Summary

- only reuse the threadlocal `main_event_loop` when it is still running
- fall back to a new loop if that threadlocal loop has stopped but not closed
- add a regression test covering the stopped-but-open loop case

A stopped loop can still accept `call_soon_threadsafe()` without ever running the scheduled task, which leaves `async_to_sync()` blocked waiting for a result.

Fixes #545.

## Testing

- `python -m pytest -q`
- `python -m mypy asgiref tests setup.py docs/conf.py`

### DIFF
--- a/asgiref/sync.py
+++ b/asgiref/sync.py
@@ -278,19 +278,21 @@ class AsyncToSync(Generic[_P, _R]):
                 finally:
                     del self.loop_thread_executors[loop]
 
-            if self.main_event_loop is not None:
+            running_in_main_event_loop = False
+            if (
+                self.main_event_loop is not None
+                and self.main_event_loop.is_running()
+            ):
                 try:
                     self.main_event_loop.call_soon_threadsafe(
                         self.main_event_loop.create_task, awaitable
                     )
                 except RuntimeError:
-                    running_in_main_event_loop = False
+                    pass
                 else:
                     running_in_main_event_loop = True
                     # Run the CurrentThreadExecutor until the future is done.
                     current_executor.run_until_future(call_result)
-            else:
-                running_in_main_event_loop = False
 
             if not running_in_main_event_loop:
                 loop_executor = None
--- a/tests/test_sync.py
+++ b/tests/test_sync.py
@@ -2,6 +2,7 @@ import asyncio
 import contextvars
 import functools
 import multiprocessing
+import os
 import sys
 import threading
 import time
@@ -15,6 +16,7 @@ import pytest
 
 from asgiref.sync import (
     AsyncSingleThreadContext,
+    SyncToAsync,
     ThreadSensitiveContext,
     async_to_sync,
     iscoroutinefunction,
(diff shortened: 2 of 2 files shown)


carltongibson:
@andersk Can I get you to look at this example? — I think it's the same issue as #528 but with a better example. 

I had similar experimenting: 

```python
def test_async_to_sync_with_stopped_main_loop():
    """
    A wrapper built while a loop is running captures that loop. If the loop is
    later stopped but not closed (e.g. a shared pytest-asyncio loop between
    tests), calling the wrapper from a sync thread must fall back to a new loop
    rather than deadlock.
    """
    holder = {}

    async def inner():
        return 42

    async def build():
        holder["fn"] = async_to_sync(inner)

    loop = asyncio.new_event_loop()
    try:
        loop.run_until_complete(build())
        assert not loop.is_running() and not loop.is_closed()

        result = {}

        def worker():
            result["value"] = holder["fn"]()

        thread = threading.Thread(target=worker, daemon=True)
        thread.start()
        thread.join(timeout=5)
        assert not thread.is_alive(), "async_to_sync deadlocked on a stopped loop"
        assert result["value"] == 42
    finally:
        loop.close()
```

It looks like the claim in #507 that the `is_running()` check was redundant was not right. 

In particular, pytest-asyncio reuses the same loop between tests with multiple `loop.run_until_complete()` invocations. (That seems legitimate.) 

WDYT? Thanks!

andersk:
I wonder if the real problem there is that that we capture the loop in `AsyncToSync.__init__` at all? It’s unintuitive that merely defining a function in a different location should affect which loop it uses when called.

If we remove that capture, your example passes, and the only failure in the test suite is the final `assert` in `test_async_to_sync_to_thread_decorator`:

https://github.com/django/asgiref/blob/4d233dd54b63c88b22470692c3d69d164a05ff92/tests/test_sync.py#L278-L279

Indeed, this is the only place in the test suite where a captured loop makes it through the top of `AsyncToSync.__call__` without being replaced.

Perhaps this test’s expectation is wrong? `asyncio.to_thread` intentionally doesn’t copy the loop to the new thread, because a loop object is not thread-safe. Granted, we only use the thread-safe subset of the loop interface. But does the user really expect that? If the user wanted to go back to the old thread, why did they write `asyncio.to_thread` instead of `sync_to_async`?

andersk:
I drafted #562 for that proposal. I’d be curious if it also prevents the pytest-asyncio deadlocks that people say they’re seeing (I mean I don’t think they’re making it up, but I haven’t seen a real reproduction posted yet).

carltongibson:
Resolved in #562. Thanks all.

davitacols:
Ok guys, I can now delete the branch

Issue #458, opened 2024-04-01
Issue in sync.py's SyncToAsync class as new ThreadPoolExecutor executors with a daemon threads getting created for requests.


sunilnitdgp:
Hi,

I'm attempting to run an async view using the ASGI protocol with Daphne as the server. However, I've noticed that it's creating new ThreadPoolExecutor instances for some requests, with daemon threads still running in the background post-benchmarking. My understanding is that since it's based on an event loop, it should use a single ThreadPoolExecutor. Could someone clarify this for me?

Middlewares:
```
MIDDLEWARE = [
    'django.middleware.security.SecurityMiddleware',
    'django.contrib.sessions.middleware.SessionMiddleware',
    'django.middleware.common.CommonMiddleware',
    'django.middleware.csrf.CsrfViewMiddleware',
    'django.contrib.auth.middleware.AuthenticationMiddleware',
    'django.contrib.messages.middleware.MessageMiddleware',
    'django.middleware.clickjacking.XFrameOptionsMiddleware',
]
```
And my view is:
```
import threading

import asyncio
import aiohttp
import logging

import psutil as psutil
from rest_framework.response import Response
from adrf.decorators import api_view
import json

logger = logging.getLogger(__name__)


@api_view(['GET'])
async def index(request):
    res = await make_api_request("http://{{host}}/v1/completions")
    return Response(res, status=200)


async def make_api_request(url, method="POST", headers=None, params=None, json_data=None, timeout=None):
    try:
    	json_data = {
                'prompt': 'Hi, How are you?',
                'max_new_tokens': 700, 'temperature': 0, 'top_p': 1, 'max_tokens': 700,
                'model': 'meta-llama/Llama-2-7b-chat-hf'}
        async with aiohttp.ClientSession() as session:
            async with session.request(method, url, headers=headers, params=params, json=json_data,
                                       timeout=timeout, ssl=False) as response:
                content = await response.read()
                if 'json' in response.headers.get('Content-Type', ''):
                    content = json.loads(content)
                return content
    except asyncio.TimeoutError:
        raise TimeoutError("Request timed out. The server did not respond within the specified timeout period.")
    except aiohttp.ClientError as e:
        raise ConnectionError(f"Request error: {str(e)}")
    except Exception as e:
        raise Exception(f"Exception error: {str(e)}")
```
The code I'm pointing to in sync.py
```
    async def __call__(self, *args: _P.args, **kwargs: _P.kwargs) -> _R:
[...]

PR #560, opened 2026-05-21
Add opt-in cancel_futures_on_exit to ThreadSensitiveContext

junodak:
## Problem

`ThreadSensitiveContext.__aexit__` calls `executor.shutdown()` with the implicit `wait=True`. When sync middleware inside the context invokes `SyncToAsync` to call back into the event loop, the request-scoped pool ends up waiting on a future the loop itself must produce:

- the main loop blocks on `executor.shutdown()`, waiting for worker threads to drain
- the worker threads block on `current_thread_executor.run_until_future`, waiting on a future the loop must run

Neither side can make progress. This is the failure mode reported in #545, #495, and #458.

## Proposal

Add an opt-in `cancel_futures_on_exit` flag that switches the shutdown call to `executor.shutdown(wait=False, cancel_futures=True)`. The default is unchanged.

Three ways to opt in:

- constructor argument
  ```python
  async with ThreadSensitiveContext(cancel_futures_on_exit=True):
      ...
  ```
- subclass attribute (useful when the context is instantiated by framework code you cannot easily reach)
  ```python
  class MyTSCtx(ThreadSensitiveContext):
      cancel_futures_on_exit = True
  ```
- environment variable for process-wide opt-in
  ```
  ASGIREF_CANCEL_FUTURES_ON_THREADSENSITIVE_EXIT=1
  ```

## Trade-off

With the flag on, pending futures are cancelled and `shutdown` returns immediately. Running worker threads cannot be interrupted mid-syscall in Python, so they finish in the background; their results are discarded because the context is already gone, and any resources they own are released on thread exit. Graceful drain is traded for liveness.

## Tests

`tests/test_sync.py` covers the historic default plus each of the three opt-in surfaces. The rest of the suite is unaffected.

## Production observation

In our deployment (Django ASGI + Daphne + ninja API, sync middleware chain underneath an async view) this deadlock manifests as the pod becoming unresponsive — no `/internal/health` response, no API requests served. Kubernetes' liveness probe eventually SIGKILLs the container, so there is no permanent outage, but every cycle costs the configured failure threshold (5 minutes in our setup) of full service unavailability. Over an 8‑day window we observed 19/19 pod restarts following this exact pattern, on average every 5.4 hours.

After applying the same change as a monkey-patch in our service, the cycle stopped under identical load. Filing this PR so users in similar situations have an upstream-blessed way out instead of patching `asgiref` in place.

## References

- #545 Nested AsyncToSync and SyncToAsync results in blocking code
- #495 sync_to_async detects deadlock in case of issue in loop.run_in_executor
- #458 ThreadPoolExecutor accumulation in SyncToAsync

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,15 @@
+Unreleased
+----------
+
+* Added an opt-in ``cancel_futures_on_exit`` flag to ``ThreadSensitiveContext``.
+  When enabled, ``__aexit__`` calls ``executor.shutdown(wait=False,
+  cancel_futures=True)`` instead of waiting for worker threads to drain — an
+  escape hatch for the cross-call deadlock that occurs when sync code in the
+  context calls back into the event loop via ``SyncToAsync``. The default is
+  unchanged. The flag can be set per-instance, per-subclass, or process-wide
+  via the ``ASGIREF_CANCEL_FUTURES_ON_THREADSENSITIVE_EXIT`` environment
+  variable.
+
 3.11.1 (2026-02-03)
 -------------------
 
--- a/asgiref/sync.py
+++ b/asgiref/sync.py
@@ -126,10 +126,31 @@ class ThreadSensitiveContext:
     >>> import time
     >>> async with ThreadSensitiveContext():
     ...     await sync_to_async(time.sleep, 1)()
+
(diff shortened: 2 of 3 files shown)


carltongibson:
Resolved in #562. Thanks all.

Issue #559, opened 2026-05-06
StatelessServer fails to start with Python 3.14 (via channels worker)


morganrallen:
Like issue #488 it appears `get_event_loop` is also being deprecated. This issue is affecting `channels` when using Python 3.14

```
root@183eb0185674:/src# python asgi_test.py 
Traceback (most recent call last):
  File "/src/asgi_test.py", line 15, in <module>
    server.run()
    ~~~~~~~~~~^^
  File "/src/asgiref/server.py", line 59, in run
    event_loop = asyncio.get_event_loop()
  File "/usr/local/lib/python3.14/asyncio/events.py", line 715, in get_event_loop
    raise RuntimeError('There is no current event loop in thread %r.'
                       % threading.current_thread().name)
RuntimeError: There is no current event loop in thread 'MainThread'.
```

asgi_test.py
```
from asgiref.server import StatelessServer


class TestServer(StatelessServer):
    def __init__(self, application, max):
        super().__init__(application, max)

    async def handle(self):
        pass

def app():
    pass

server = TestServer(app, 1)
server.run()
```

It also looks like `tests/test_server.py` doesn't test the `.run` method.

My quick fix to keep on development has been changing https://github.com/django/asgiref/blob/main/asgiref/server.py#L59 to
```
try:
  event_loop = asyncio.get_event_loop()
except:
  event_loop = asyncio.new_event_loop()
```

PERSONAL OPINION: Its passed time to get off GitHub, codeberg is real friendly and has far better uptimes, no AI scraping our every word and no awful React frontend that barely works on a good day.

morganrallen:
Being deprecated appears to be incorrect, as noted here https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop, it seems that the behavior has changed. At a guess, it used to create a loop if one didn't exist? Now it notes it will throw the `RuntimeError` I'm seeing above. It also suggests using `asyncio.run` might be the correct way.

carltongibson:
@morganrallen It looks like we're missing test coverage here. If you fancy adding a case, no problem with updating to use the modern API.

carltongibson:
Fixed in #561

PR #561, opened 2026-05-21
Use asyncio.run in StatelessServer.run

puneetdixit200:
`StatelessServer.run()` currently calls `asyncio.get_event_loop()` before running `arun()`. On Python 3.14, a fresh main thread may not have a current event loop, so this raises before the server can start.

This switches the synchronous entry point to `asyncio.run(self.arun())`, which creates and manages the loop for this one-shot runner. I also added a regression test that patches `asyncio.get_event_loop()` to raise with the Python 3.14-style failure and confirms `run()` still reaches the handler.

Refs #559.

Tests:

- `.venv\Scripts\python -m pytest tests\test_server.py::test_run_creates_event_loop_when_none_exists -q`
- `.venv\Scripts\python -m pytest tests\test_server.py -q`
- `.venv\Scripts\python -m pytest -q`
- `.venv\Scripts\python -m mypy .`
- `git diff --check`

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -7,6 +7,10 @@ UNRELEASED
 
 * Dropped support for EOL Python 3.9.
 
+* Fixed ``StatelessServer.run()`` failing on Python 3.14, where
+  ``asyncio.get_event_loop()`` no longer creates an event loop if none
+  exists. It now uses ``asyncio.run()``. (#559)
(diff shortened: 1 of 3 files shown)


carltongibson (approved):
Thanks @puneetdixit200. I adjusted the test and added a release note. 🎁

PR #562, opened 2026-05-22
 Don’t capture the event loop in `AsyncToSync.__init__`

andersk:
It was unintuitive that merely defining a function in a different location should affect which loop it uses when called.

`test_async_to_sync_to_thread_decorator` was the only test in the whole test suite where a captured loop made it through the top of `AsyncToSync.__call__` without being replaced. This test was wrong to expect us to capture the loop across `asyncio.to_thread`:

`asyncio.to_thread` intentionally doesn’t copy the running loop to the new thread, because a loop object is not thread-safe. Granted, we only use the thread-safe subset of the loop interface. However, the user shouldn’t expect that. If they wanted to go back to the old thread, they would have written `sync_to_async` instead of `asyncio.to_thread`.

Includes the test case from https://github.com/django/asgiref/pull/558#issuecomment-4515994185, which now passes.

### DIFF
--- a/asgiref/sync.py
+++ b/asgiref/sync.py
@@ -207,17 +207,12 @@ class AsyncToSync(Generic[_P, _R]):
         except AttributeError:
             pass
         self.force_new_loop = force_new_loop
-        self.main_event_loop = None
-        try:
-            self.main_event_loop = asyncio.get_running_loop()
-        except RuntimeError:
-            # There's no event loop in this thread.
-            pass
 
     def __call__(self, *args: _P.args, **kwargs: _P.kwargs) -> _R:
         __traceback_hide__ = True  # noqa: F841
 
-        if not self.force_new_loop and not self.main_event_loop:
+        main_event_loop = None
+        if not self.force_new_loop:
             # There's no event loop in this thread. Look for the threadlocal if
             # we're inside SyncToAsync
             main_event_loop_pid = getattr(
@@ -226,7 +221,7 @@ class AsyncToSync(Generic[_P, _R]):
             # We make sure the parent loop is from the same process - if
             # they've forked, this is not going to be valid any more (#194)
             if main_event_loop_pid and main_event_loop_pid == os.getpid():
-                self.main_event_loop = getattr(
+                main_event_loop = getattr(
                     SyncToAsync.threadlocal, "main_event_loop", None
                 )
 
@@ -284,10 +279,10 @@ class AsyncToSync(Generic[_P, _R]):
                 finally:
                     del self.loop_thread_executors[loop]
 
-            if self.main_event_loop is not None:
+            if main_event_loop is not None:
                 try:
-                    self.main_event_loop.call_soon_threadsafe(
-                        self.main_event_loop.create_task, awaitable
+                    main_event_loop.call_soon_threadsafe(
+                        main_event_loop.create_task, awaitable
                     )
                 except RuntimeError:
                     running_in_main_event_loop = False
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,9 +1,33 @@
 UNRELEASED
 ----------
 
-* Fixed a deadlock of the entire event loop when exiting
-  ``ThreadSensitiveContext`` while its executor thread was still blocked
-  waiting on the event loop. (#535)
+* ``AsyncToSync`` no longer captures the running event loop on
+  instantiation. (#562)
+
+  This resolves a series of deadlocks that users experienced after asgiref 3.9.0,
+  particularly with pytest-asyncio. pytest-asyncio stops the event loop between
+  tests, and long-running unawaited futures could find themselves trying to
+  schedule work onto a stopped loop, and so would never complete. Ideally, code
+  should be structured to await long-running futures before returning, but this
+  change should help users experiencing issues here.
+
+  The loop is now resolved when the callable is invoked rather than when it is
+  created. If ``async_to_sync`` is called from within ``sync_to_async``, the
+  parent event loop is still used, as before.
+
+  The possibility of deadlock therefore remains in some nested patterns. For
+  example, an async function may call a long-running ``sync_to_async`` function
+  that itself uses ``async_to_sync``; if the outer function returns before the
+  sync future completes, the parent event loop may already be stopped, and the
+  nested calls cannot be driven to completion.
+
+  This is not a bug in asgiref — the same patterns deadlock in plain asyncio. As
+  above, restructure your code to await the ``sync_to_async`` future before
+  exiting the driving coroutine.
+
+* Fixed an event loop deadlock when exiting ``ThreadSensitiveContext``
+  while its executor thread was still blocked waiting on the event loop.
+  (#535)
 
 * Dropped support for EOL Python 3.9.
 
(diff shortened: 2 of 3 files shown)


asgiref/sync.py
```diff
@@ -220,7 +215,7 @@ def __call__(self, *args: _P.args, **kwargs: _P.kwargs) -> _R:
             # We make sure the parent loop is from the same process - if
             # they've forked, this is not going to be valid any more (#194)
             if main_event_loop_pid and main_event_loop_pid == os.getpid():
-                self.main_event_loop = getattr(
+                main_event_loop = getattr(
```
carltongibson, line 224: I wonder if there's a version of the deadlock that could come up along this branch, still requiring the `is_running()` check 🤔


Update: See 327c9661ef9d873b655a40d96ebcbbc137f48d46 to assess.

carltongibson (commented):
This looks interesting 👀 

There'll be a small behaviour change right, which we can clarify about when to use `to_thread` vs `sync_to_async` (and `thread_sensitive`)

carltongibson:
Hey @andersk — I added 327c9661ef9d873b655a40d96ebcbbc137f48d46 to show the concern along threadlocal loop path — can you let me know what you think?

carltongibson:
Just fyi, I'm at PyCon Italia this week and intended to try and resolve the set of issues here and get a release out. My current thinking is to take your patch but also handle the extra case shown by the test I added (as separate commits). If you have any thoughts do let me know! Thanks again for the input!

andersk:
Looking at your new test:

`inner()` runs in the main thread if the loop is still running (e.g. if `asyncio.sleep(0.1)` is increased to `asyncio.sleep(0.5)`), but your patch allows it to run in a new thread otherwise. It still seems dangerous to have code that may or may not run on the main thread depending on timing; this is a recipe for introducing thread safety bugs that are difficult to reproduce and understand.

Is your test fundamentally different from the following deadlock that doesn’t use `asgiref` at all? Why should adding a round trip through `async_to_sync` and `sync_to_async` shift the burden of avoiding the deadlock onto asgiref?

```py
import asyncio
import threading

finished = threading.Event()

async def inner():
    await asyncio.sleep(0.3)
    finished.set()

async def driver():
    asyncio.ensure_future(inner())
    await asyncio.sleep(0.1)

loop = asyncio.new_event_loop()
loop.run_until_complete(driver())
assert not loop.is_running() and not loop.is_closed()
finished.wait()
```

Maybe it would help to see the actual pytest-asyncio example you’re investigating, if you have one.

carltongibson:
It's more that we're introducing a change in behaviour from what we've always done here. (In the face of the reported deadlocks.) This one code path is all that remains of the possible "stopped loop" case. So... what's the correct behaviour at that point?

spreeni:
This seems sound and minimal and solves my observed stalling tests 🙌  Will be happy to see this adopted as it is currently preventing us from migrating to Django 6.x

carltongibson:
Thanks for commenting @spreeni. Are you able to confirm whether the stall is resolved for you without https://github.com/django/asgiref/pull/562/commits/327c9661ef9d873b655a40d96ebcbbc137f48d46 or only with?

andersk:
As a reminder, I’m still interested in looking at a reproducible example of a deadlock with pytest-asyncio and investigating whether we can find a real solution that doesn’t introduce thread safety issues like 327c9661ef9d873b655a40d96ebcbbc137f48d46 does, but nobody has posted one.

carltongibson:
@andersk I think what we have here is already showing it 🤔

This is more or less the same thing spread over two tests: 

```
import threading
import time

import pytest

_finished = threading.Event()
_state = {}


async def inner():
    return "inner-ok"


def blocking():
    # On an asgiref sync_to_async worker thread. threadlocal.main_event_loop is
    # set (by thread_handler) to the module loop that dispatched us.
    time.sleep(0.5)  # outlive the test's run_until_complete -> loop stops under us
    _state["value"] = async_to_sync(inner)()  # routes onto the stopped module loop
    _finished.set()


@pytest.mark.asyncio(loop_scope="module")
async def test_a_launch_worker():
    import asyncio

    # Fire-and-forget: start the worker but don't await it, so this coroutine
    # returns (and the loop stops) while the worker is still going.
    asyncio.ensure_future(sync_to_async(blocking)())
    await asyncio.sleep(0.05)  # let the worker get dispatched and started


def test_worker_completes():
    # Plain sync test. While this runs the module loop is stopped-but-open, so the
    # worker's async_to_sync call has nothing driving its target loop.
    ok = _finished.wait(timeout=6)
    assert ok, "DEADLOCK: async_to_sync hung on the threadlocal stopped loop"
    assert _state["value"] == "inner-ok"
```

It's the same problem: trying to run a nested async function on a stopped loop. (Why the nesting? 🫠 — But I think it's something folks do.) 

My query is really what's best for the user? 

* We used to fallback to a new loop, via the old is_running check. That might not be _right_ but it seems better than...
* Deadlock: what we have now. 
* Option 3? Could we, say, error loudly here? 🤔

Something like this, far too literal but…, maybe: 

```
  if main_event_loop is not None and not main_event_loop.is_running():
      raise RuntimeError(
          "async_to_sync() was called from a sync_to_async() worker whose event "
          "loop is no longer running. The loop that dispatched this thread has "
          "stopped, so the awaitable cannot be scheduled back onto it. This usually "
          "means the awaiting coroutine returned or was cancelled while the worker "
          "was still in flight (e.g. a fire-and-forget sync_to_async() call). Await "
          "the sync_to_async() call to completion, or pass force_new_loop=True if "
          "you intend to run on a fresh loop."
      )
```

WDYT?

carltongibson:
Hmm. Option 3 breaks several tests (at first pass). Can't tell between the _broken_ nested case, and just a plain _stale_ loop. 😕

andersk:
That’s another clear case of “user wrote code whose behavior is to deadlock, and is shocked when we do what they told us to do”. If you do exactly the same test without `asgiref` involved, you get exactly the same deadlock.

```python
import asyncio
import threading
import pytest

_finished = threading.Event()
_state = {}

async def inner():
    return "inner-ok"

async def non_blocking():
    await asyncio.sleep(0.5)
    _state["value"] = await inner()
    _finished.set()

@pytest.mark.asyncio(loop_scope="module")
async def test_a_launch_worker():
    asyncio.ensure_future(non_blocking())
    await asyncio.sleep(0.05)

def test_worker_completes():
    ok = _finished.wait(timeout=6)
    assert ok, "DEADLOCK"
    assert _state["value"] == "inner-ok"
```

If we can find a way for `asgiref` to prove that a deadlock has occurred in some cases, then sure, throwing an error on a best-effort basis would be nice. I suspect that’s hard here, because we can’t be sure that the user isn’t going to start running the event loop again. Maybe a warning would work if we think this case is rare.

But, if we have to choose between faithfully obeying an instruction to deadlock and magically doing the wrong thing, such as arbitrarily switching to a new thread that the user didn’t ask for and that might lead to thread unsafety and data corruption, then **deadlocking is the correct choice**.

I’m still open to looking at a more convincing test case, but it’s starting to seem likely that the problem really is with the tests and not with `asgiref`.

carltongibson:
> If we can find a way for asgiref to prove that a deadlock has occurred in some cases, then sure, throwing an error on a best-effort basis would be nice. I suspect that’s hard here

Yes, that's the way (if there is one). 

Thanks for the back and forth @andersk — I agree with you, but it helps to push against the options. 

I'll pull in your fix here, and leave the is_running check out:: if folks are **still** hitting deadlocks after that, at least we have a clear _"This is what you're almost certainly doing"_ story to give them.

carltongibson (approved):
I added a release note. 

Thanks for this @andersk 🎁

PR #565, opened 2026-06-19
Wrap async_to_sync with functools.update_wrapper

### LINKED ISSUE #500: `async_to_sync' is not wrapped properly:
For now, I'm using this approach, which should be in the `asgiref`:
```python
import functools
...

from asgiref.sync import async_to_sync as _async_to_sync
...

def async_to_sync(awaitable: Callable[P, Coroutine[Any, Any, T]]) -> Callable[P, T]:
    return functools.wraps(awaitable)(_async_to_sync(awaitable))
```

patrickwehbe:
Fixes #500.

`async_to_sync` doesn't copy the wrapped function's metadata onto the callable it returns, so attributes like `__name__` and `__wrapped__` are missing. People work around it by wrapping the result in `functools.wraps` themselves (the issue shows this).

`sync_to_async` already calls `functools.update_wrapper(self, func)` in its `__init__`, and `AsyncToSync.__get__` does the same when used as a method decorator. `AsyncToSync.__init__` was the only one that didn't, so this was just an inconsistency.

The fix is one line in `AsyncToSync.__init__`. I also added a test next to the existing `sync_to_async` attribute test. Without the fix the new test fails with `AttributeError: 'AsyncToSync' object has no attribute '__name__'`; with it the name, docstring, `__wrapped__` and custom attributes all carry over.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,10 @@
+UNRELEASED
+----------
+
+* ``async_to_sync`` now copies the wrapped function's metadata (``__name__``,
+  ``__wrapped__``, etc.) onto the returned callable, matching ``sync_to_async``.
+  (#500)
+
 3.11.1 (2026-02-03)
 -------------------
 
--- a/asgiref/sync.py
+++ b/asgiref/sync.py
@@ -196,6 +196,7 @@ class AsyncToSync(Generic[_P, _R]):
                 "async_to_sync was passed a non-async-marked callable", stacklevel=2
             )
         self.awaitable = awaitable
+        functools.update_wrapper(self, awaitable)
         try:
             self.__self__ = self.awaitable.__self__  # type: ignore[union-attr]
         except AttributeError:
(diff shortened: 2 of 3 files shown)


carltongibson (commented):
Hey @patrickwehbe — I think we need to handle functions marked with `markcoroutinefunction()` here. 

 Could you scrub the markers after copying:

```
functools.update_wrapper(self, awaitable)
# update_wrapper copies __dict__, which carries the markcoroutinefunction()
# marker for marked callables — this callable is sync, so remove it.
self.__dict__.pop("_is_coroutine_marker", None)
self.__dict__.pop("_is_coroutine", None)  # Before Python 3.12. 
```

…and extend the test to assert not iscoroutinefunction(wrapped), using a marked input (or sync_to_async(f)) — a plain `async  def` has an empty __dict__ so won't exercise this.

Django marks async CBVs and middleware this way, so this is a realistic path
  — e.g. the round-trip `sync_to_async(async_to_sync(sync_to_async(f)))` is going to cause issues here otherwise.

Issue #504, opened 2025-04-11
timeout does not call uncancel() after supressing cancellation


graingert:
asgiref.timeout.timeout does not call uncancel()

PR #568, opened 2026-06-29
Use asyncio.timeout on Python 3.11+

8rief:
## Summary

- use `asyncio.timeout` for `ApplicationCommunicator` on Python 3.11+
- keep the existing vendored timeout unchanged as the Python 3.10 fallback
- cover cancellation-count restoration after a handled `receive_output` timeout

Related to #504.

## Verification

The regression test failed before the production change because the handled timeout left the current task's cancellation count at `1` instead of restoring it to `0`.

- `pytest tests/test_testing.py -q`
- `pytest -q` — 98 passed, 1 xfailed on Python 3.12
- `mypy .`
- `pre-commit run --files asgiref/testing.py tests/test_testing.py`
- Python 3.10 container — 95 passed, 2 skipped, 1 xfailed; mypy passed
- Python 3.11 and 3.14 containers — focused tests passed
- Python 3.14 container — 98 passed, 1 xfailed; mypy passed

### DIFF
--- a/asgiref/testing.py
+++ b/asgiref/testing.py
@@ -1,9 +1,14 @@
 import asyncio
 import contextvars
+import sys
 import time
 
 from .compatibility import guarantee_single_callable
-from .timeout import timeout as async_timeout
+
+if sys.version_info >= (3, 11):
+    async_timeout = asyncio.timeout
+else:
+    from .timeout import timeout as async_timeout
 
 
 class ApplicationCommunicator:
--- a/tests/test_testing.py
+++ b/tests/test_testing.py
@@ -48,6 +48,32 @@ async def test_receive_nothing():
     assert await instance.receive_nothing(0.01) is True
 
 
+@pytest.mark.asyncio
+async def test_receive_output_timeout_restores_cancellation_count():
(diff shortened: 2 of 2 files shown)


carltongibson:
@graingert WDYT?

carltongibson:
We're dropping Python 3.10 very shortly. I wonder if the correct approach here isn't to switch testing.py to `asyncio.timeout` instead, and deprecate timeout.py here. (We might merge this anyway...)

carltongibson:
Needs a rebase.

graingert:
I'd be very tempted to use the built in asyncio.timeout where available and leave this timeout alone for 3.10

8rief:
Rebased and revised this along the suggested compatibility boundary. `ApplicationCommunicator` now uses `asyncio.timeout` on Python 3.11+, while Python 3.10 keeps the existing vendored fallback unchanged.

I moved the regression coverage to `ApplicationCommunicator.receive_output`; it confirms that a handled timeout no longer leaves the caller task in a cancelling state. The full test suite and mypy pass on Python 3.10 and 3.14, with focused checks on Python 3.11 as well.

PR #572, opened 2026-07-14
Restore previous SyncToAsync.__call__ internal code shape.

### LINKED ISSUE #571: SyncToAsync change in 3.12.0 leaks New Relic's injected kwarg into the wrapped function (TypeError on every Django ASGI request)
## Summary

3.12.0 changed `SyncToAsync.__call__` so the callable handed to `thread_handler` is no longer the `functools.partial(self.func, *args, **kwargs)` itself, but a `child()` closure wrapping it (deda0d4, #564). The New Relic Python agent depends on the previous shape: it injects a `_nr_current_trace` kwarg in its `SyncToAsync.__call__` wrapper and pops it back off `partial.keywords` in its `SyncToAsync.thread_handler` wrapper. With 3.12.0 the pop fails silently, the kwarg reaches the wrapped user function, and any callable that doesn't accept arbitrary kwargs raises.

On Django this means every request through the ASGI handler fails with a 500: the first `sync_to_async` call per request is `request_started.asend`, whose target is a local closure:

```
TypeError: Signal.asend.<locals>.sync_send() got an unexpected keyword argument '_nr_current_trace'
```

## Environment

- asgiref 3.12.0 (same app works on 3.11.1)
- Django 5.2.7 (ASGI, uvicorn), Python 3.11
- newrelic 13.2.0 (agent enabled)

## Reproduction

Run any Django ASGI app under the New Relic agent with asgiref 3.12.0:

```bash
pip install Django==5.2.7 uvicorn newrelic==13.2.0 asgiref==3.12.0
NEW_RELIC_LICENSE_KEY=... newrelic-admin run-program uvicorn myproject.asgi:application
curl localhost:8000/anything   # -> 500
```

## Analysis

3.11.1, `SyncToAsync.__call__`:

```python
child = functools.partial(self.func, *args, **kwargs)
func = context.run
```

3.12.0:

```python
inner = functools.partial(self.func, *args, **kwargs)

def child() -> _R:
    _restore_context(context)
    return inner()
```

New Relic's instrumentation ([`newrelic/hooks/adapter_asgiref.py`](https://github.com/newrelic/newrelic-python-agent/blob/main/newrelic/hooks/adapter_asgiref.py)):

```python
async def wrap_SyncToAsync__call__(wrapped, instance, args, kwargs):
    kwargs["_nr_current_trace"] = current_trace()
    return await wrapped(*args, **kwargs)


def thread_handler_wrapper(wrapped, instance, args, kwargs):
    try:
        bound_args = bind_args(wrapped, args, kwargs)
        trace = bound_args["args"][0].keywords.pop("_nr_current_trace", None)
    except Exception:
        trace = None
    ...
```

In 3.12.0, `bound_args["args"][0]` is the `child` closure, which has no `.keywords`, so the `except` swallows the `AttributeError` and `_nr_current_trace` stays in `inner.keywords` and is forwarded to the target function.

Truncated traceback:

```
File ".../asgiref/sync.py", line 521, in __call__
    ret = await asyncio.shield(exec_coro)
File ".../concurrent/futures/thread.py", line 58, in run
File ".../newrelic/hooks/adapter_asgiref.py", line 34, in thread_handler_wrapper
    return wrapped(*args, **kwargs)
File ".../asgiref/sync.py", line 576, in thread_handler
    return func(*args, **kwargs)
File ".../asgiref/sync.py", line 502, in child
    return inner()
TypeError: Signal.asend.<locals>.sync_send() got an unexpected keyword argument '_nr_current_trace'
```

## Notes

I realize this is arguably New Relic's bug — their hook reaches into asgiref internals that were never a public contract, and it likely needs a fix on their side regardless. Filing here as a datapoint: 3.12.0 currently breaks every Django + New Relic ASGI deployment that installs with an unpinned asgiref, so you may want to weigh keeping the partial shape, or at least a mention in the changelog.

Workaround: pin `asgiref==3.11.1`.

carltongibson:
Ensure child partial is the is the deferred sync function to be run, with its args and kwargs bound. Fixes #571.


phongthanfz:
Hi @carltongibson,

We tested this PR in our project (Django 5.2.7 ASGI + New Relic 13.2.0 + uvicorn) and confirmed it fixes the regression from #571.

- `asgiref==3.12.0` + `newrelic-admin run-program`: fails with `TypeError: sync_function() got an unexpected keyword argument '_nr_current_trace'`.
- `asgiref @ this PR branch (a503183)` + `newrelic-admin run-program`: works correctly.

Restoring the internal `functools.partial` shape and adding the canary test look good. From our side, this is good to merge as a hotfix.

Thanks for the quick fix.

PR #575, opened 2026-07-18
Remove NGINX Unit

rhoboro:
I removed NGINX Unit.
Because [NGINX Unit's official site](https://unit.nginx.org/ ) announced "As of October 2025, NGINX Unit is archived and unmaintained." and the [nginx/unit](https://github.com/nginx/unit) repository has also been archived.


carltongibson (commented):
I think it's probably better to remove this entirely.

rhoboro:
@carltongibson Thank you for your reply. I did it.

carltongibson (approved):
Great. Thank you.

PR #586, opened 2026-08-30
Allow HTTP/3 as an http_version value

### LINKED ISSUE #544: HTTP/3 official support
I think we need to change the wording on the `HTTP` section, and add "3" as possible value to `http_version`.

gyanu2507:
The HTTP spec still listed http_version as 1.0/1.1/2 even though pseudo-headers already talk about HTTP/3 and several implementations already send "3".

Bumped the www spec to 2.6 and updated the intro paragraphs so HTTP/3 is a first-class transport, same as HTTP/2.

Fixes #544

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,9 @@
+UNRELEASED
+----------
+
+* Document HTTP/3 as an allowed ``http_version`` value in the ASGI HTTP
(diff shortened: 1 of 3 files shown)


gyanu2507:
HTTP/3 is already in main via #589 (merged 2026-09-16), which also closed #544. This branch now conflicts with that wording, so I am closing it as a duplicate.

Issue #466, opened 2024-08-18
ASGI TLS extension


Kludex:
There's a PR on uvicorn to add this extension: https://github.com/encode/uvicorn/pull/1119

And I have a question... The author said we either need to maintain the list that is maintained by IANA, or use `cryptography` as dependency.

On the spec, this is written:

```
client_cert_name duplicates information that is also available in client_cert_chain. However, many ASGI applications will probably find that information is sufficient for their application - it provides a simple string that identifies the user. It is simpler to use than parsing the x509 certificate. For the server, this information is readily available.

There are theoretical interoperability problems with client_cert_name, since it depends on a list of object ID names that is maintained by IANA and theoretically can change. In practice, this is not a real problem, since the object IDs that are actually used in certificates have not changed in many years. So in practice it will be fine.
```

As a maintainer of uvicorn, I don't really want to generate the list as the PR is doing, neither add `cryptography` as dependency. Suggestions? Was this spec created by theory, or it was actually implemented when written?

andrewgodwin:
The spec was contributed by someone else so I'm afraid I can't speak to the intent, nor did I implement it.

What's the specific technical problem - client_cert_name can't be generated without a list of IDs from some external source?

Kludex:
> What's the specific technical problem - client_cert_name can't be generated without a list of IDs from some external source?

Yes. It looks like I can use `cryptography` package to help here, but is the spec forcing me to?

Just to confirm, is my understanding correct @synodriver? (they implemented this extension on `nonecorn`)

jschlyter:
The extension specifies that the `cipher_suite` should be _a 16-bit unsigned integer that encodes the pair of 8-bit integers specified in the relevant RFC, in network byte order_. This is information that is available in the TLS stack (e.g., OpenSSL), but is not exposed. To resolve this one has to down the [Transport Layer Security (TLS) Parameters](https://www.iana.org/assignments/tls-parameters/tls-parameters.xml) from IANA and generate the list of name to integer, or use a library that can provide this mapping. Also, this integer means nothing to developers nor users, so to use one has to convert it back to a cipher name.

IMHO, it would be a lot simpler if the `cipher_suite` was just a string with the SSL object cipher name (e.g., `TLS_AES_256_GCM_SHA384`).

jschlyter:
For the `client_cert_name` I would opt for just dropping it for the reasons above.

mdgilene:
[...]

PR #590, opened 2026-09-15
Prefer client certificate chains in the TLS extension

Kludex:
## Summary

- Keep only `client_cert_chain` in the TLS extension.
- Remove fields that require servers to parse or translate TLS metadata.
- Document the TLS extension version history.

This addresses the implementation concerns in #466. Applications can derive certificate details from the chain without requiring ASGI servers to parse X.509 certificates.

## Testing

- `git diff --check`
- `uv run --with sphinx --with sphinx-rtd-theme sphinx-build -E -b html docs docs/_build/html`

## AI Disclaimer

This PR was developed with the assistance of either Claude or Codex. I've reviewed and verified the changes.


sylvesterkaczmarek (commented):
This removes fields that 0.2 made mandatory (`server_cert`, `tls_version`, and `cipher_suite`), but the TLS extension carried in `scope` has no version marker or negotiation mechanism: applications only know that `scope['extensions']['tls']` exists. An existing 0.2 consumer can therefore receive a 0.3 scope and fail when those required keys disappear. Could the existing fields remain available/optional for compatibility, or could an explicit extension version be carried before removing them?

Kludex:
We can keep the `spec_version`.

sylvesterkaczmarek:
Yes, keeping an explicit spec_version in the TLS extension payload and bumping it for 0.3 would address the compatibility concern. That gives existing 0.2 consumers a way to distinguish the new shape before relying on fields that 0.3 removes. Happy with that direction.

sylvesterkaczmarek (approved):
Rechecked d5b0c8d8. The TLS scope now carries spec_version="0.3", with an explicit compatibility rule that a missing marker means 0.2. That gives existing consumers a reliable way to distinguish the field set before reading keys removed in 0.3. My compatibility concern is resolved.

